PR #168, opened 2015-01-12
Snapshot api

lebauce:


smira:
I'm going to work on merging that...

smira:
My work plan:

 * snapshot diff should be GET, not POST
 * snapshot creation from package refs might take sources somehow (names?), this is low-priority
 * do we need pull/merge? (low-priority)
 * review system tests
 * unify "package search" code path

smira:
And most important: thanks a lot @lebauce for your great contribution!

smira:
@lebauce I've changed `?format=details` output to be really verbose JSON with all the package fields exposed, along with aptly-specific `Key`, `Short-Key` and `FilesHash`

Issue #116, opened 2014-10-06
REST API for aptly


smira:
MVP scenario:

* upload package files over HTTP
* add package to local repo from file (uploaded or local)
* publish/update local repo

Good to have:

* complete local repo management
* basic snapshot management
* full publishing support

lebauce:
@smira Thanks for the review !

Ok, I'll lock the collection if sorting is requested. I'm not sure either if the sorting should be done in the client side. Maybe doing this in the server side will be more useful if paging is implemented.

Regarding apiSnapshotsCreate, are you talking about the "SourceIDs" argument ? If so, my understanding was that the 'SourceIDs' of the Snapshot structure was meant to keep track of the originating snapshot/mirror/repository. Therefore apiSnapshotsCreate accepts a SourceIDs array and checks that the sources exist. Should I fetch the package list of the sources and merge it with the PackageRefs argument ?

smira:
Some items left:

 * publish API - to completion
 * periodically flush CollectionFactory to GC unused objects
 * API docs (huge!)
 * package show API

smira:
Items left:

 * publish API: list, drop, switch
 * API docs (in progress)

sepulworld:
I'd like to test out some of the API functions in the nightly release on aptly-vagrant.  Are you testing API functionality through "aptly serve"?  Or through a proxy like Nginx? Curious what you testing setup looks like.

sepulworld:
Ok, I see now:

aptly api serve -listen=:8081

smira:
This is documentation for new version, work in progress: http://beta.aptly.info/doc/api/

smira:
Finally it is done. Pushing 0.9 out!!!!

kumy:
Great news!
I have to test that :)

vitalyisaev2:
I have an urgent need in REST API... Using reprepro at the moment, but wanna to move to aptly. I'm going to test it now. Thanks for a nice work.

PR #169, opened 2015-01-12
Graph REST API #116

lebauce:


smira:
@lebauce thanks a lot!

I've done some refactoring and enabled custom format for graphics.

PR #174, opened 2015-01-21
Implemented apiPublishUpdateSwitch.

mkoval:
I implemented `apiPublishUpdateSwitch` for updating repositories through the REST API. This has been working fine for me over several days of relatively heavy testing. Most of this logic is copied verbatim from the command-line `aptly repo update` call. Ideally the common code could be factored out into a standalone update repository function that is used in both places.

I only made this change myself because I needed the ability to update a published (local) repository through the REST API and couldn't wait for it to be implemented. You're welcome to ignore this if you have a better implementation in mind.

This is the first time I have ever written Go, so no promises on the quality. :smile:

### DIFF
--- a/api/publish.go
+++ b/api/publish.go
@@ -168,7 +168,67 @@ func apiPublishRepoOrSnapshot(c *gin.Context) {
 
 // PUT /publish/:prefix/:distribution
 func apiPublishUpdateSwitch(c *gin.Context) {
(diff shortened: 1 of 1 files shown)


smira:
Thanks a lot, @mkoval! I would merge this as soon as I get done with snapshot API.

PR #201, opened 2015-02-12
d-i requires the Suite field in order to validate a mirror.

alexanderguy:
Debian's installer validates a mirror by downloading a Release,
and then cross-checking it based on its Codename and Suite.  Without
a Suite field, the installer becomes unhappy (e.g. segfaults) and
won't continue the install.

Making the Codename and Suite the same validates with no problem.

### DIFF
--- a/deb/publish.go
+++ b/deb/publish.go
@@ -624,6 +624,7 @@ func (p *PublishedRepo) Publish(packagePool aptly.PackagePool, publishedStorageP
 	release := make(Stanza)
 	release["Origin"] = p.GetOrigin()
 	release["Label"] = p.GetLabel()
+	release["Suite"] = p.Distribution
 	release["Codename"] = p.Distribution
 	release["Date"] = time.Now().UTC().Format("Mon, 2 Jan 2006 15:04:05 MST")
 	release["Architectures"] = strings.Join(utils.StrSlicesSubstract(p.Architectures, []string{"source"}), " ")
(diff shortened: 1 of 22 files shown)


alexanderguy:
Ah, I thought I had run all the tests.  I'll fix the tests, and update the branch.

smira:
Thanks a lot!

rmedaer:
Thank's @alexanderguy,

I got exactly the same issue by writing a custom distro (Debian derived) with aptly as APT mirror.

For those would like to know the reason why:
`d-i/cdrom-detect package` detects automatically the `distribution` by reading the `suite` field in `Release` file. If this field is not set, it reject the installation:

_From "debian/cdrom-detect.postinst" :_
```
# Set the suite and codename used by base-installer and base-config
# to the suite/codename that is on the CD. In case there are multiple
# suites, prefer the one in default-release.
set_suite_and_codename() {
        for dir in $(cat /etc/default-release) $(ls -1 /cdrom/dists/); do
                relfile="/cdrom/dists/$dir/Release"
                if [ -e $relfile ]; then
                        suite=$(sed -n 's/^Suite: *//p' "$relfile")
                        codename=$(sed -n 's/^Codename: *//p' "$relfile")
                        log "Detected CD with '$suite' ($codename) distribution"
                        db_set cdrom/suite "$suite"
                        db_set cdrom/codename "$codename"

                        break
                fi
        done
}

# (...)

set_suite_and_codename

if [ -z "$suite" ]; then
        log "Error reading Release file; unable to determine distribution"
        db_input critical cdrom-detect/no-release || [ $? -eq 30 ]
        db_go
        fail
fi
```

`cdrom/suite` should be used by `d-i/base-installer` but ... not really not ! **However**, it's used by `d-i/apt-setup` in generators.

PR #218, opened 2015-03-07
Add basic support for non-Amazon S3

srvg:
This PR is meant for review only. `make test` is passing, I think I updated everything that was expected.

FYI

At this point I only extended the config with the S3endpoint parameter from aws.Region
This keeps by default `S3LowercaseBucket` to true, while I tested a bucket name with some upper case letters, which works on this backend. Perhaps this feature would benefit from looking at more parameters? Or in a future patch...

Now on the patch itself.

Tested it with http://auroraobjects.eu/ (S3 on top of Ceph backend) shows publishing repo's work as expected.

However dropping a published repo, yields errors, strangely a different one depending of using tls or not.

Without tls (plain http):
```
$ aptly publish drop squeeze s3:aptly:aptly
ERROR: unable to remove: error deleting multiple paths from S3: auroradam:aptly/: 411 Length Required
```
With https:
```
$ aptly publish drop squeeze s3:aptly:aptly
ERROR: unable to remove: error deleting multiple paths from S3: auroradam:aptly/: Request Content-Type is not multipart/form-data
```

At this point, I'm not sure where to look.

### DIFF
--- a/context/context.go
+++ b/context/context.go
@@ -322,7 +322,7 @@ func (context *AptlyContext) GetPublishedStorage(name string) aptly.PublishedSto
 
 			var err error
 			publishedStorage, err = s3.NewPublishedStorage(params.AccessKeyID, params.SecretAccessKey,
-				params.Region, params.Bucket, params.ACL, params.Prefix, params.StorageClass,
+				params.Region, params.S3endpoint, params.Bucket, params.ACL, params.Prefix, params.StorageClass,
 				params.EncryptionMethod, params.PlusWorkaround)
 			if err != nil {
 				Fatal(err)
--- a/man/aptly.1
+++ b/man/aptly.1
@@ -1,7 +1,7 @@
 .\" generated with Ronn/v0.7.3
 .\" http://github.com/rtomayko/ronn/tree/0.7.3
 .
-.TH "APTLY" "1" "February 2015" "" ""
+.TH "APTLY" "1" "March 2015" "" ""
 .
 .SH "NAME"
 \fBaptly\fR \- Debian repository management tool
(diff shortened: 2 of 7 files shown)


srvg:
BTW, shouldn't `man/aptly.1` be added to .gitignore?

smira:
I wonder if auroraojbects does support Multiple Delete API? If not, we can workaround that using "single delete".

In `s3/public.go`, method `RemoveDirs`, you can change `const page` to 1 (one) and change `storage.bucket.MultiDel(paths)` to `storage.bucket.Del(paths[0])`.

srvg:
You are correct. With the patch you described, it works.
I can't find any reference to multi-object deletion in Ceph docs (http://ceph.com/docs/master/radosgw/s3/objectops/)

srvg:
This behaviour is Ceph Object Gateway S3 APi - specific. Do we need a way to keep track of such differences, depending on the backend - if the user can  know that.. I incidently know this because I know the company behinds auroraobjects...

smira:
I believe we can make this config option: does this "S3 provider" support `MultiDel` or not - and choose correct implementation. `MultiDel` with 1000 objects per request would be several orders of magnitude faster I believe when deleting huge published repositories.

srvg:
I can imagine in the future there might be other S3 backends, with other specifics. Should we take that into account and perhaps think on how to structure something more generic, in the code and in the config file, to support future exceptions?

smira:
@srvg perhaps first option on MultiDel support should be enough for now, we can figure out later what we need. I can take over this patch to merge into aptly if you don't mind. I would need some testing from you for the final version

srvg:
I had been thinking, perhaps we can define a 'backend type' for S3. In this case, that would be Ceph-someversion, and that would imply no MultiDel?

I don't mind you taking over. I'm also willing to give it a try. I was waiting on your feedback :)

smira:
@srvg I've implemented custom S3 endpoint based on your pull request. Could you please try it out?

It is available as nightly build `aptly_0.9.6~dev+91+g1c8f151`.

srvg:
(still on my todo list, too busy atm)

smira:
@srvg I'm closing this, please feel free to re-open...

PR #220, opened 2015-03-15
Remove prefix directory on publish drop

### LINKED ISSUE #198: Aptly leave empty directory structure when droping published repo with prefix
When removing the last `distribution` from a published repo with a prefix, Aplty will empty directories, but leave the prefix structure in place.

phenomenes:
Fix #198


phenomenes:
Sorry, I'll look at the tests when I have some time

PR #231, opened 2015-03-31
Update import.go

seaninspace:
Add support for Automatic Debug Packages (.ddeb's)


smira:
@seaninspace does it work for you?

PR #232, opened 2015-04-06
Remove empty Depends from published metadata

shanegibbs:
This fix for #233 works in theory, however, is this the right place to do this? and is this the right thing to do? - I'll look at tests depending on the answer to these questions.


smira:
I believe we should remove such fields while parsing control file. I'll do better fix.

PR #242, opened 2015-04-24
aptly graph specific output filename

gdbdzgd:
https://github.com/smira/aptly/issues/241

	modify：     cmd/graph.go

Signed-off-by: Zhang, Guodong <gdzhang@linx-info.com>


smira:
Thanks!

Issue #261, opened 2015-06-09
miss '\n' in publish Sources


gdbdzgd:
missing '\n'  at first line in multilineFields

public/dists/jessie/main/source/Sources

need:
```
Files:
 3a91f06cafa02b06912d6cd3b24e0903 78732 bash_4.3-11.debian.tar.xz
 505b3a2c129ad52cc53288bfda702457 1540 bash_4.3-11.dsc
 557827017514aaa0d1e3888da7567cfc 7516231 bash_4.3.orig.tar.gz
Vcs-Browser: https://code.launchpad.net/~doko/+junk/pkg-bash-debian
Package-List: bash deb shells required arch=any
 bash-builtins deb utils optional arch=any
 bash-doc deb doc optional arch=all
 bash-static deb shells optional arch=any
Checksums-Sha1:
 1d761a503adb5769a2233a566f055d0c4c62f383 78732 bash_4.3-11.debian.tar.xz
 625e4b3194aa2a21dbc28133225f28e3ff835b61 1540 bash_4.3-11.dsc
 a44778703fa80ba13baa9134f9086ad09cc5147a 7516231 bash_4.3.orig.tar.gz
Vcs-Bzr: http://bazaar.launchpad.net/~doko/+junk/pkg-bash-debian
Homepage: http://tiswww.case.edu/php/chet/bash/bashtop.html
Checksums-Sha256:
 e042d5be5df8df339a516290bb27b1cdf027cbba0bd6c695e62f6782b4d72dba 78732 bash_4.3-11.debian.tar.xz
 81394eb07c09c244e5d01c6be19e69fc60153575f11923cb10ea1ed87b71ce5b 1540 bash_4.3-11.dsc
 b2fe79ddf9e7abdb4695e3070afa866d2a94a58d1cc9d731585330c753815491 7516231 bash_4.3.orig.tar.gz
```

but now is:

```
Files: 3a91f06cafa02b06912d6cd3b24e0903 78732 bash_4.3-11.debian.tar.xz
 505b3a2c129ad52cc53288bfda702457 1540 bash_4.3-11.dsc
 557827017514aaa0d1e3888da7567cfc 7516231 bash_4.3.orig.tar.gz
Vcs-Browser: https://code.launchpad.net/~doko/+junk/pkg-bash-debian
Package-List: bash deb shells required arch=any
 bash-builtins deb utils optional arch=any
 bash-doc deb doc optional arch=all
 bash-static deb shells optional arch=any
[...]

PR #262, opened 2015-06-09
Add missing '\n' at first line in multilineFields

gdbdzgd:
public/dists/[distribution]/main/source/Sources
	missing '\n' at first line in multilineFields

	modify：     deb/package.go

Signed-off-by: Zhang, Guodong <gdzhang@linx-info.com>


smira:
Thanks for the PR, I've provided better fix in #261.

PR #263, opened 2015-06-09
Fix various Gin API changes in tip

dullgiulio:
In case you want to keep up with the latest Gin, I think these changes are all needed. I also updated the Gomfile accordingly.

I wrongly tried to install via go-get and got compliation errors, this is why I noticed.


dullgiulio:
Unsure why the build failed. Is Gom updating dependencies?

smira:
@dullgiulio `Gomfile` update is not included in this changeset.

dullgiulio:
Sorry, you are right. I included it now.

PR #266, opened 2015-06-19
fix EOF during mirror update

sobczyk:
see http://stackoverflow.com/a/19006050

### DIFF
--- a/http/download.go
+++ b/http/download.go
@@ -145,6 +145,7 @@ func (downloader *downloaderImpl) handleTask(task *downloadTask) {
 		task.result <- fmt.Errorf("%s: %s", task.url, err)
 		return
 	}
+	req.Close = true
 
 	proxyURL, _ := downloader.client.Transport.(*http.Transport).Proxy(req)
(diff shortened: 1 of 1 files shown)


smira:
@sobczyk doesn't it effectively disable keep-alive?

sobczyk:
@smira I'm not sure, I'm a golang newbie and this was a copypaste fix

Before `aptly mirror update` of ubuntu repositories failed often, making it useless in a daily cron script. reprepro works fine

I guess this could be a CLI switch.

smira:
So it seems that Go is thinking that it's okay to leave connection in the pool and re-use it for the next request, while actually connection has been closed at the server-side. If we add `req.Close = true`, every connection would be closed (no keep-alive, no connection pool), but this problem goes away. I wonder how can we workaround that...

sobczyk:
I'll link the golang bug https://github.com/golang/go/issues/4677

I experienced failures even with the `req.Close = true` a while ago...

smira:
Thanks, @sobczyk! 

Probably it's better to have `req.Close = true` before it gets fixed upstream. Aptly is downloading big files in several execution threads, so keep-alive shouldn't affect overall download speed that much.

PR #270, opened 2015-07-08
fix some packages not being included, because of badly parsed folded field

sobczyk:
according to https://www.debian.org/doc/debian-policy/ch-controlfields.html
Binary filed is a folded type.

During building boost genchanges folds Binary to multiple lines. (it has many binary packages from one source)
When aptly trims whitespace it then contatenates package names at the end of the line and beginning of the next line


smira:
@sobczyk thanks for discovering this bug! I've fixed it in a bit different way, which should be still compatible (whitespace is preserved)

smira:
Fix should be available as nightly build `aptly_0.9.6~dev+145+g76bf7cb`

PR #300, opened 2015-09-07
Add a flag to unlock database after each API request

### LINKED ISSUE #234: Database locked while running api serve
While running aptly's api service, commands such as `mirror` don't work as the api service locks the database. Is this expected behaviour?

vincentbernat:
After the first API request, the database was locked as long as the API
server is running. This prevents a user to also use the command-line
client. This commit adds a new flag `-no-lock` that will close the
database after each API request.

Closes #234

### DIFF
--- a/cmd/api_serve.go
+++ b/cmd/api_serve.go
@@ -46,6 +46,7 @@ Example:
 	}
 
 	cmd.Flag.String("listen", ":8080", "host:port for HTTP listening")
+	cmd.Flag.Bool("no-lock", false, "don't lock the database")
 
 	return cmd
 
--- a/api/api.go
+++ b/api/api.go
@@ -32,25 +32,27 @@ func cacheFlusher() {
 	for {
 		<-ticker
 
-		// lock everything to eliminate in-progress calls
-		r := context.CollectionFactory().RemoteRepoCollection()
-		r.Lock()
-		defer r.Unlock()
+		func() {
+			// lock everything to eliminate in-progress calls
+			r := context.CollectionFactory().RemoteRepoCollection()
+			r.Lock()
+			defer r.Unlock()
 
-		l := context.CollectionFactory().LocalRepoCollection()
-		l.Lock()
-		defer l.Unlock()
+			l := context.CollectionFactory().LocalRepoCollection()
+			l.Lock()
+			defer l.Unlock()
 
-		s := context.CollectionFactory().SnapshotCollection()
-		s.Lock()
-		defer s.Unlock()
+			s := context.CollectionFactory().SnapshotCollection()
+			s.Lock()
+			defer s.Unlock()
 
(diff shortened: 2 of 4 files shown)


vincentbernat:
Just pushed a second version of the commit...

vincentbernat:
Hummm, this works randomly in fact.

aioue:
+1

vincentbernat:
It's easy to make the code panic with multiple concurrent HTTP requests due to the fact that a concurrent request can close the database. However, it also fails to work after some time and just concurrent requests. I'll try to propose another approach (mediation through a goroutine for example).

vincentbernat:
Again, a new version using goroutines. This time, I didn't find an obvious way to make it crash. I leave it running for a few hours to check if my problems are solved.

vincentbernat:
New version, using a defer to ensure that database is unlocked even if there is some panic.

vincentbernat:
Well, for some reason, this is worse than the previous version... I have added some debug code to help me find why it goes wrong.

vincentbernat:
Another update. It seems that my problem was I didn't account for the cacheFlusher goroutine.

vincentbernat:
With the latest iteration, the locking problem is gone from my point of view. It has been running for 3 days without any problem.

mattes:
+1

cmd/api_serve.go
```diff
@@ -46,6 +46,7 @@ Example:
 	}
 
 	cmd.Flag.String("listen", ":8080", "host:port for HTTP listening")
+	cmd.Flag.Bool("no-lock", false, "don't lock the database")
```
mattes: I think this should be `true` per default.

avongluck-r1soft:
Whoo! Is this one ready to merge?  This locking mess is the #1 issue that drives us nuts :-)

bjoernHeneka:
+1

pztrick:
:+1:

pztrick:
I believe there is another issue. It seems to me that if I perform the following it will fail.

1. Make an API call
2. Create a snapshot via CLI
3. Make an API call using newly created snapshot name

Restarting the API daemon will allow it to see the new snapshot created via CLI.

I believe the API daemon must keep state of DB in process (which of course is why there should DB lock). Is there way to also re-open DB with each new API call so that it would pull in newly created snapshot? How expensive is this?

smira:
@pztrick yep, that's different issue... There's `cacheFlusher` already which flushes all cached data in API instance: https://github.com/smira/aptly/blob/master/api/api.go#L29 

In ideal world, DB lock should be dropped after all cached data is flushed

smira:
Patch looks really good, thanks! I need to look into it deeply.

cmd/api_serve.go
smira: To be on the safe side, we can leave it as `false`?

PR #307, opened 2015-10-02
Fix lock handling in cache flusher for API

vincentbernat:
Unlocking the different elements in cache flusher was deferred to the
end of the function. Unfortunately, being a for loop wrapped in a
goroutine, deferred were never executed.

I think that unlocking was still done due to garbage collection.


smira:
:+1: thank you!

Issue #264, opened 2015-06-12
enhancement: daemonized API service


aioue:
aptly should install and start its API service automatically in the background on install and boot like a standard linux service (nginx etc).

Configuration should be stored in /etc/ and working files (DB etc) in its own directory.

[![Bountysource](https://www.bountysource.com/badge/issue?issue_id=21733238)](https://www.bountysource.com/issues/21733238-enhancement-daemonized-api-service?utm_source=21733238&utm_medium=shield&utm_campaign=ISSUE_BADGE)

AlekSi:
It should be easy with https://github.com/fhd/init-script-template

oCCtet:
Automatically... not until #234 is resolved, as api serve apparently makes cli use impossible.

I guess one possibility would be to have an option in /etc/default/aptly, which - if set to a value like `AUTO_API_SERVE=1` - would cause aptly initscript to start the service on boot.

aioue:
As long as the service installs and runs and acts a lot more like a 'regular' service, how it's accomplished is broadly irrelevant.

I'd expect to see maybe an 'aptly' user created by the package, an `/etc/aptly` directory for config, and the DB data kept in it's own directory, something like `/usr/local/aptly`.

#234 isn't a blocker. You have to tell a service like haproxy to be enabled in the config before it'll run automatically. When the CLI eventually moves to calling the API under the covers, then even that requirement will go away.

oCCtet:
Originally aptly was not limited to one-instance-per-machine; for example, each user could maintain their own package repositories (think PPAs...). The proposal of automatically installing aptly as a machine-wide service strongly suggests a severely limited use case.

IMHO making the CLI use API service as the backend is a terrible idea. It should be the other way around: higher-level interfaces should use lower-level functions, i.e. HTTP interface should use command-line utilities (or library calls if such were available) to perform its tasks.

aioue:
Running as a service is no different from multiple people trying to bind to the same default port using `aptly api serve`. I'm only repeating smira's ideas in #234 re: CLI calling API since he's the owner of the project and ultimately decides its direction.

@AlekSi if you can get it functioning as it is, but working as a persistent service via some init scripts, the bounty is yours.

aioue:
This looks promising:
"A daemon package for use with Go (golang) services with no dependencies."
https://github.com/takama/daemon

igorastds:
[...]

PR #312, opened 2015-10-08
enhancement: daemonized API service #264

fabrepe:
- Daemonize and packaging scripts
- copy file from orphaned branche #264


smira:
@fabrepe why is it required?

fabrepe:
This is destined to add the daemonize scripts. 
It also provide packaging scripts necessary to set up aptly as a service (issue 264)

smira:
Packaging scripts are part of Debian distribution, I don't see why they should be part of aptly repo. If you want to change Debian packaging, please work with Debian maintainer. If you wan to do your own packaging - please feel free to use your own repo.

PR #320, opened 2015-11-25
add diagnostic output

paul-krohn:
Let the user know which deb file is corrupt or badly formatted when throwing an error. No functional change.


smira:
:+1: thanks, @paul-krohn

PR #324, opened 2015-12-02
[SECURITY] Only download Files without checksum when checksum is ignored anyway

x539:
Downloading and using e.g. Packege.bz2 when there is no Checksum for it
will break the trust-chain started from the Release-file.


smira:
@x539 thanks for the PR, I've committed the fix and now I need to fix unit-tests as well

PR #344, opened 2016-02-03
Use official AWS SDK; support STS credentials

### LINKED ISSUE #342: S3 upload does not support a session token
Aptly seems to ignore `AWS_SESSION_TOKEN` when authenticating against AWS, and as such, can't be used with assumed roles, EC2 instance profiles, or any other kind of STS credentials. Such things are used by organizations that have multiple AWS accounts. For example I use [awsudo](https://github.com/paperg/awsudo) to assume a role in a test account or a production account as appropriate. Or I might be running aptly on an EC2 instance which is assigned a role which can access the S3 bucket hosting my repo.

See https://blogs.aws.amazon.com/security/post/Tx3D6U6WSFGOK2H/A-New-and-Standardized-Way-to-Manage-Credentials-in-the-AWS-SDKs for the standard credential-finding logic supported by all official AWS SDKs. I've had to fix similar issues in other projects, and usually the solution is to not pass any credentials at all to the SDK. It will then find credentials using standard methods.

### LINKED ISSUE #343: Locking issues, api calls non-functional after cli
When running the apty 9.6.0 api in no-lock, api calls like graph.png break after cli aptly executes.

[GIN] 2016/02/03 - 13:05:20 | 200 | 1.991951807s | 127.0.0.1:55292 |   GET     /api/graph.png
run CLI tools
[GIN] 2016/02/03 - 13:16:04 | 500 | 169.975462ms | 127.0.0.1:55359 |   GET     /api/graph.png
[GIN] 2016/02/03 - 13:16:07 | 500 |  183.16334ms | 127.0.0.1:55292 |   GET     /api/graph.png
[GIN] 2016/02/03 - 13:16:08 | 500 |  180.58619ms | 127.0.0.1:55292 |   GET     /api/graph.png
[GIN] 2016/02/03 - 13:16:15 | 500 | 170.654224ms | 127.0.0.1:55292 |   GET     /api/graph.png

bitglue:
Now that there's an official Go AWS SDK from Amazon, use that instead of
goamz. goamz isn't getting much love these days.

Implement support for STS credentials, as in assumed roles and EC2
instance profiles. The configuration is extended to support a session
token, though I'm not sure why anyone would put temporary credentials in
a configuration file. More likely, no credentials will be explicitly
configured at all, and they will be discovered through the standard SDK
mechanisms described at
<https://blogs.aws.amazon.com/security/post/Tx3D6U6WSFGOK2H/A-New-and-Standardized-Way-to-Manage-Credentials-in-the-AWS-SDKs>.

Resolves #342.


smira:
thank you, this looks really great, I'm going to run some tests with it

bitglue:
Cool. I was just trying to figure out how to get the test suite running locally, so I didn't have to wait for Travis each time.

smira:
@bitglue locally you need to set up AWS creds, this should be enough to enable it

smira:
I've created branch `bitglue-aws_sdk` with some integration fixes

bitglue:
I wasn't able to figure out the tests. It looks like goamz had some kind of test stub, but I don't think there's any such thing for the AWS SDK.

However I have been using a local build with these changes and it's working for my purposes.

smira:
I want to re-use s3test from goamz for unit-tests if possible. If not, we will drop unit-tests for S3 publishing and stay only with functional tests

smira:
Pushed new version with unit-tests, still some work left, but looks much better now.

smira:
`aptly_0.9.7~dev+11+gfcedaa3_amd64` is a nightly build which includes this PR

bitglue:
Thanks for accepting the patch and fixing the tests! I've been using the nightly build today and it seems to be working great.

smira:
Thanks for great PR! I did only one change to force "path-style" S3 requests if custom endpoint is provided. It should match previous behaviour (with goamz)

smira:
One more PR related to that: #355

PR #346, opened 2016-02-11
Flush collection contents on each DB unlock in API.

smira:
See #343

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -27,6 +27,29 @@ const (
 	RELEASEDB
 )
 
+// Flushes all collections which cache in-memory objects
+func flushColections() {
+	// lock everything to eliminate in-progress calls
+	r := context.CollectionFactory().RemoteRepoCollection()
+	r.Lock()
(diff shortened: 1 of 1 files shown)


smira:
@vincentbernat could you please take a look?

api/api.go
```diff
@@ -39,38 +62,12 @@ func cacheFlusher(requests chan int, acks chan error) {
...
-				defer func() {
-					requests <- RELEASEDB
-					<-acks
-				}()
+			// if aptly API runs in -no-lock mode,
+			// caches are flushed when DB is closed anyway, no need
+			// to flush them
+			if requests == nil {
+				flushColections()
 			}
```
vincentbernat: No need to keep enclosing anonymous function since we don't need a scope anymore.

vincentbernat:
LGTM, but not tested.

api/api.go
smira: yeah, thanks, I'll fix it

PR #366, opened 2016-03-17
gpg: Sign with SHA-256 for compatibility with APT 1.2.7

geofft:
APT 1.2.7, which was released earlier this week and will be in Ubuntu 16.04 LTS, gives a warning on signed repositories where the GPG signature uses SHA-1 or weaker. This will be upgraded to a hard error in Ubuntu 16.10 and in some post-release update to 16.04, as well as in Debian 9. See the [developer's blog post](https://juliank.wordpress.com/2016/03/14/dropping-sha-1-support-in-apt/) and [a follow-up post](https://juliank.wordpress.com/2016/03/15/clarifications-and-updates-on-apt-sha1/) for more information.

This change forces aptly to sign Release and InRelease files with SHA-256. gpg's default signature algorithm is SHA-1, so e.g. http://repo.aptly.info/dists/squeeze/InRelease uses SHA-1. (Users will need to re-publish all existing repositories after this change.)

### DIFF
--- a/utils/gpg.go
+++ b/utils/gpg.go
@@ -143,7 +143,7 @@ func (g *GpgSigner) Init() error {
 func (g *GpgSigner) DetachedSign(source string, destination string) error {
 	fmt.Printf("Signing file '%s' with gpg, please enter your passphrase when prompted:\n", filepath.Base(source))
 
-	args := []string{"-o", destination, "--armor", "--yes"}
+	args := []string{"-o", destination, "--digest-algo", "SHA256", "--armor", "--yes"}
 	args = append(args, g.gpgArgs()...)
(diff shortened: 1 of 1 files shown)


smira:
Nice, thank you!

thebwt:
Any chance to get this merged into some kind of release before 16.04 drops? At this point every apt command returns this warning :( even a v0.9.6.1 would help with end-user exp.

smira:
I think we can do v0.9.7... let me see if I can do that next week

geofft:
@thebwt: You can probably put `digest-algo sha256` into your `~/.gnupg/gpg.conf` as a workaround.

smira:
0.9.7 has been released

PR #367, opened 2016-03-18
Fixed mkdir mode from 755 to 777

bentoi:
Directory creation isn't using the 0777 mode as advertised by the FAQ, this fixes the mode in few places (there might be other places where using 0777 is appropriate).


smira:
Thank you!

PR #396, opened 2016-05-16
Allow overriding architecture info from Release file

dstelter:
Adds a flag -force-architectures to ignore missing architectures from
mirrors. This flag can be used in cases where the mirrored repository
does not provide an "Architecture: " line.

Example Release file:
http://mitaka-jessie.pkgs.mirantis.com/debian/dists/jessie-mitaka-backports/Release


smira:
👍 thank you!

PR #397, opened 2016-05-17
gh_issue/388_api_hook

### LINKED ISSUE #388: FeatureRequest: Add hooks
TL;DR:
I'd like to have an option to add hooks when certain things are done, in my special case I'd like to have some hook for an update of a published repo.

I have aplty's API running behind an nginx and use that for pushing some packages from the CI system to aplty and potentially update a local repo. However I'd like to e.g. synchronize that repo to different regions and therefore would like to have aptly call out (to either an API or to a local command).
I don't want to push that to the CI system, as I have multiple ways how packages get pushed to aptly; I want it to be the responsibility of aplty to call out when "something" happened. Basically in the end I want to have a possibility to specify some hook for each(?) action exposed via API. I am happy when this is a single command in the aptly.conf which then needs to do the "routing" and is run as the aptly user.

hoegaarden:
Enables one to add hooks for API calls as requested in #388.

closes #388

### DIFF
--- a/api/router.go
+++ b/api/router.go
@@ -7,31 +7,34 @@ import (
 	"strconv"
 	"strings"
 	"os/exec"
+	"encoding/json"
 )
 
 var context *ctx.AptlyContext
 
 // middleware to track API calls and call a hook script
-func ApiHooks(cmd string) gin.HandlerFunc {
+func ApiHooks(cmd string, conf string) gin.HandlerFunc {
 	return func(c *gin.Context) {
 		c.Next()
 
-		// env := os.Environ()
 		var env []string
-		env = append(env, ("METHOD="      + c.Request.Method))
-		env = append(env, ("REQ_URL="     + c.Request.URL.String()))
-		env = append(env, ("REMOTE_HOST=" + c.Request.RemoteAddr))
-		env = append(env, ("STATUS="      + strconv.Itoa(c.Writer.Status())))
+		env = append(env,
+			("APTLY_METHOD="      + c.Request.Method),
+			("APTLY_REQ_URL="     + c.Request.URL.String()),
+			("APTLY_REMOTE_ADDR=" + c.Request.RemoteAddr),
+			("APTLY_STATUS="      + strconv.Itoa(c.Writer.Status())),
+			("APTLY_CONFIG="      + conf),
+		)
 
 		// collect all named params
 		for _, p := range c.Params {
-			env = append(env, "PARAM_" + p.Key + "=" + p.Value)
+			env = append(env, "APTLY_PARAM_" + p.Key + "=" + p.Value)
 		}
 
 		// collect the query params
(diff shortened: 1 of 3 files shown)


hoegaarden:
Any update / input on that? Is there anything I can do to get this or something like that merged?

aol-nnov:
@hoegaarden thank you for the contribution! It is a really nice feature and I need it too! :)
But unfortunately I'm not familiar with go. Could you please rebase your work to the tip of current mainline? (Adding some documentation or usage example would be a great plus too!)
It seems @smira is way too busy to look into this project (which is very, very sad and disappointing) and he hasn't found any maintainer yet..

api/router.go
```diff
@@ -4,17 +4,55 @@ import (
...
+			("APTLY_PARAMS="      + string(params)),
+			("APTLY_QUERY="       + string(query)),
+		}
+
+		cmd := exec.Command(cmd)
+		cmd.Env = env
+
+		// fire and forget
+		go func(){
+			cmd.Run()
```
smira: I think logging errors is required here, otherwise failures to run hook script would go completely unnoticed?

api/router.go
```diff
@@ -4,17 +4,55 @@ import (
...
 // Router returns prebuilt with routes http.Handler
 func Router(c *ctx.AptlyContext) http.Handler {
 	context = c
 
 	router := gin.Default()
 	router.Use(gin.ErrorLogger())
 
+	conf := context.Config()
+	api_hook_cmd := conf.APIHookCmd
+	if api_hook_cmd != "" {
```
smira: Go is using `camelCase` (`apiHookCmd`), same with `confStr`

utils/config.go
```diff
@@ -24,6 +24,7 @@ type ConfigStructure struct {
 	SkipContentsPublishing bool                        `json:"skipContentsPublishing"`
 	S3PublishRoots         map[string]S3PublishRoot    `json:"S3PublishEndpoints"`
 	SwiftPublishRoots      map[string]SwiftPublishRoot `json:"SwiftPublishEndpoints"`
+	APIHookCmd             string                      `json:"APIHookCmd"`
```
smira: This breaks system tests, so simple fix is required to make tests happy again...

api/router.go
```diff
@@ -4,17 +4,55 @@ import (
...
 
+// middleware to track API calls and call a hook script
+func ApiHooks(cmd string, conf string) gin.HandlerFunc {
+	return func(c *gin.Context) {
+		c.Next()
+
+		params , _ := json.Marshal(c.Params)
+		query, _   := json.Marshal(c.Request.URL.Query())
+
+		env := []string{
```
smira: Do we want to inherit from the aptly env here?

sliverc:
I am closing this PR in favor of #597 so we can focus our effort there.

PR #406, opened 2016-05-30
API authentication

mgrdevport:
@smira 
Thank you for making aptly, I really enjoy using it.

For secure api access I've included [gin-jwt](https://github.com/appleboy/gin-jwt) and implemented the basic authorization and authentication pattern.
### New config file content:
- apiSecretKey
  (string containing the api secret key)
- apiUsers
  (dict containing username, password and array of roles)

Example:

```
...
"apiSecretKey": "supersecretkey",
"apiUsers": {  
    "admin": {
        "password": "admin",
        "roles": ["admin"]
    }
},
...
```
### Modified system test:
- config file creation with new content elements
- added parameter _--auth_ to use authentication on api system test
- run api system test with or without authentication (see Makefile)
### Modified Makefile:
- new task _system-test.api_ (standalone api authentication system test)
- new task _system-test.api.noauth (standalone api system test without authentication - default)

Question:
Should we test the api with and without authentication when running default system test? We could do so if we add the task _system-test.api_ to the default sytem-test or all.

Anonymous access without authentication is still granted if api users are not defined in the config file (backwards compatibility).

Currently there is only support for the role admin.
### Security note:

Aptly config file should be read protected (plain passwords)
Secure api access with SSL to prevent security issues (Nginx, Apache, ...)


smira:
Thanks, this looks really interesting. I was looking more towards protecting API with nginx and basic auth if required, but this brings out of the box solution which might be good enough.

mgrdevport:
The documentation should recommend the usage of nginx/apache as SSL proxy. I am planning to use the api in conjunction with a continuous delivery server to update mirrors, create snapshots, publish tehm and as preparation for web frontends.

The return code for the SIGINT termination (t07_serve) of the process seems to be different when using go 1.5 / go 1.6 in the travis ci builds (1.5 -> 2, 1.6 -> **-**2). 
The python documentation says that the return code for the signal N would be -N (at least on Unix).

morph027:
https://morph027.gitlab.io/post/protect-aptly-api-with-basic-authentication/ :)

morph027:
Also this ;)

https://morph027.gitlab.io/post/push-deb-packages-from-ci-jobs-to-aptly-repo/

If you'd like, i can take care of the documentation for nginx/ssl setup here (once i figured out your gitlab-pages structure *g*)

smira:
@morph027 any docs are really welcome as PRs to https://github.com/aptly-dev/aptly-dev.github.io repository

morph027:
Sure, will have a look at it tomorrow!

sevki:
has this been rejected?

Issue #408, opened 2016-06-03
Feature Request: Expose Metrics


triggity:
in order to better monitor aptly , should it have a way of exposing metrics about itself. at say a /metrics endpoint. 

https://godoc.org/github.com/prometheus/client_golang/prometheus is a good candidate. 

what do you think?

devKlausS:
would be great!

PR #410, opened 2016-06-10
 #408 adding basic prometheus instrumentation

triggity:
creating basic instrumentation. 

The library imported I'm not teh biggest fan of. It basically re-does the same named prometheus client functions (https://godoc.org/github.com/prometheus/client_golang/prometheus#InstrumentHandlerFunc) with the gin context.

metrics exposed at /metrics


smira:
I'm perfectly fine to add some level of instrumentation, but I think we might want to make it more configurable and documented.

PR #412, opened 2016-06-10
Feature/description

kkupciunas:
Currently, **Description** field in published repository Release files is a hardcoded string "_Generated by aptly_" - allow specifying custom description when publishing by specifying command line option
`-description`


apachelogger:
It would be lovely if you could rebase this on master. It may also benefit from a system test.

brennerm:
What's the issue preventing this from being merged?

smira:
@brennerm system test would be really nice

dlitz:
It would be good if we could customize _any_ field in the Release file, not just the specific ones anticipated by the aptly developer.  For example, I'd like to make a backports-style repo, which would mean adding `NotAutomatic: yes` and `ButAutomaticUpgrades: yes` to the Release file.

PR #416, opened 2016-06-16
initial sftp publishing support

apachelogger:
presently still has the following problems:
- doesn't do md5 summing (which we can't really do tbh, file size compare
  and maybe mtime comparision might be an option, but that's about it)
- ssh keys cannot be specified
- possibly should allow running arbitrary pre-/post- publish commands to
  prepare the remote? e.g. publishing and then moving the full tree to
  an actual public directory to avoid transactional file inconsistencies
  (e.g. there is a time frame during which the repo has no Release file
  because of sftp limitations)

I'd love some input on this.

In particular I could use some input on how to prevent PublishedStorage from leaking sftp/ssh clients. PublishedStorage should kind of hold the two during its life time but since PublishedStorage has no Close we ultimately leak the connections since they never get closed I think.


sftp/public.go
```diff
@@ -0,0 +1,373 @@
...
+func (s *sftpClient) CreatePath(path string) (sftpFileInterface, error) {
+	// Call into our sftp.Client, coerce it's type into an interface type for
+	// testing.
+	return s.Create(path)
+}
+
+type sshClientInterface interface {
+}
+
+// PublishedStorage abstract file system with published files (actually hosted on S3)
```
l1k: s/S3/sftp/

sftp/public.go
```diff
@@ -0,0 +1,373 @@
...
+
+	// verify that package pool is local pool in filesystem
+	_ = sourcePool.(*files.PackagePool)
+
+	baseName := filepath.Base(sourcePath)
+	relPath := filepath.Join(publishedDirectory, baseName)
+	poolPath := storage.expandPath(relPath)
+	fmt.Fprintf(os.Stderr, "LinkFromPool basename:%s relpath:%s poolpath:%s\n", baseName, relPath, poolPath)
+
+	// FIXME: we do not do md5 validation or anything!
```
l1k: The [sftp spec](https://filezilla-project.org/specs/draft-ietf-secsh-filexfer-02.txt) has no provisions for retrieving the hash of a given remote file, so the only option I see is to retrieve the entire file over the network and compare it to the local file in the pool.

l1k: I've added functionality to compare the MD5 hash of a remote file in commit l1k/aptly@ffc5cbfef5d028b4a65af4079090bf6da0898cee, feel free to leverage that for your use case. As you've suggested, this could be augmented with an option to only compare metadata (timestamp, size) to achieve decent performance on a low-bandwidth connection.

apachelogger:
I have no time to pursue this, unfortunately.

PR #417, opened 2016-06-17
Print current source names (snapshot list & show)

winged:
When renaming a snapshot that was referenced by another via snapshot merge,
then the references are not updated. Consider the following:

```
$ aptly snapshot create foo from repo xyz
$ aptly snapshot merge bar foo
$ aptly snapshot rename foo foo3
$ aptly snapshot list
List of snapshots:
 * [foo3]: Snapshot from repo...
 * [bar]: Merged from sources: 'foo'
```

As you can see, after renaming the "source" snapshot, it's stored description
may refer to a snapshot that does no longer exist.

This change always generates the description from the actual source snapshots,
so it will point to the right (current) snapshot name, even after renames.


winged:
So it turns out that my version was rather incomplete. My colleague @sliverc built a much cleaner implementation in #472 so I think we should roll with that instead.

PR #418, opened 2016-06-23
update unit tests

btkostner:
- I updated the data in the mirror files so unit tests pass
- Updated api serve exit code in tests to `-2`. As far as I know aptly is still working as intended so I'm not sure if this was changed upstream or if the unit test is actually failing. Let me know if this is incorrect.


btkostner:
I now realize the answer to number `2`. Go 1.5 exits with `2` and 1.6 exists with `-2`. I guess the best way to deal with this for now is if it ends with `2` or `-2` until aptly drops go 1.5.

josegonzalez:
👍 This looks like a positive change :)

smira:
👍 thanks!

Issue #384, opened 2016-04-18
generate checksums for all Components files [$15]


btkostner:
It would be amazing if aptly would generate [AppStream](https://www.freedesktop.org/wiki/Distributions/AppStream/) data when it publishes a repository. This would allow packages in the published repo to be seen by new-age app centers, including gnome software.

[Here](https://github.com/ximion/appstream-generator) is a generator written in D for reference.

ximion:
As long as Aptly reads the `dep11/` directory and puts the correct checksums for `Components-*` into the `InRelease` file, asgen and Aptly should work well together.
I would advise against writing an own metadata generator, since that is a huge pain, and keeping it up-to-date is an issue. But making asgen easier to use from other tools would be possible (at time, it is mainly built for large distro repositories and to be a standalone tool).

btkostner:
I'm changing this issue's scope to generate checksums for all `Components` files. This will be the easiest solution to this issues, and some other issues related to metadata generation.

#### Expected behavior:
- Have a script that generates `Components` files from the aptly repository.
- Link the folder so you end up with a route similar to `/aptly/public/stable/dists/xenial/main/dep11/Components-arm64.yml.gz`
- When aptly switches published repo or snapshot, it includes checksums for all `Components` files in `/aptly/public/stable/dists/xenial/main`

btkostner:
I also put a $15 bounty on this for anyone who want's to add it:
https://www.bountysource.com/issues/33086807-generate-checksums-for-all-components-files

ximion:
Just something to add before it gets forgotten: The same thing needs to be done with the `icons-*` tarballs in the same `dep11` subdirectory.
Also, the checksums for the uncompressed files need to be in `InRelease`, but the actual uncompressed files need not (and probably should not) be present.

danrabbit:
Bumped the bounty to a total of $50 :)

apachelogger:
I have been looking this, alas, it is a bit tricky. I'd absolutely love some input on this @smira 

We need a bunch of information which could be different per source/publishedrepo/usecase.
Namely the following appstream-generator attributes:
- `UseIconTheme` (overrides icon to use)
- `MediaBaseURL` (fully qualified URI to a media directory)
- `HTMLBaseURL` (URL to a html directory, if I am not mistaken this can actually be relative)
- `DataPriority` (allows overriding of other appstream data)
The other attributes we can pretty much fill with information we already have in aptly anyway.

[...]

PR #420, opened 2016-06-27
add basic appstream support

btkostner:
So this is a basic implementation for appstream support #384
This is my first golang code so please forgive any bad practices.

This expects you to have a file path similar to the setup repository with uncompressed appstream files like such: `/var/appstream/stable/dists/xenial/main/Components-amd64.yml`. You would then set the `appStreamDir` field in the aptly configuration file to `/var/appstream`. On publish aptly would then compress the files as needed, and use them in Release file.

More improvements:
- use the compressed files outputted by asgen instead of requiring the uncompressed version.

### DIFF
--- a/Gomfile
+++ b/Gomfile
@@ -22,6 +22,7 @@ gom 'github.com/smira/lzma', :commit => '7f0af6269940baa2c938fabe73e0d7ba4120568
 gom 'github.com/golang/snappy', :commit => '723cc1e459b8eea2dea4583200fd60757d40097a'
 gom 'github.com/syndtr/goleveldb/leveldb', :commit => '917f41c560270110ceb73c5b38be2a9127387071'
 gom 'github.com/ugorji/go/codec', :commit => '71c2886f5a673a35f909803f38ece5810165097b'
+gom 'github.com/ulikunitz/xz', :commit => 'f852725cf3ed4e7dca9bd458fe881ff02b27eae4'
 gom 'github.com/vaughan0/go-ini', :commit => 'a98ad7ee00ec53921f08832bc06ecf7fd600e6a1'
 gom 'github.com/wsxiaoys/terminal/color', :commit => '5668e431776a7957528361f90ce828266c69ed08'
 gom 'golang.org/x/crypto/ssh/terminal', :commit => 'a7ead6ddf06233883deca151dffaef2effbf498f'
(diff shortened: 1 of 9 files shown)


Gomfile
```diff
@@ -22,6 +22,7 @@ gom 'github.com/smira/lzma', :commit => '7f0af6269940baa2c938fabe73e0d7ba4120568
 gom 'github.com/golang/snappy', :commit => '723cc1e459b8eea2dea4583200fd60757d40097a'
 gom 'github.com/syndtr/goleveldb/leveldb', :commit => '917f41c560270110ceb73c5b38be2a9127387071'
 gom 'github.com/ugorji/go/codec', :commit => '71c2886f5a673a35f909803f38ece5810165097b'
+gom 'github.com/ulikunitz/xz', :commit => 'f852725cf3ed4e7dca9bd458fe881ff02b27eae4'
```
smira: This library says it's "alpha quality", at the same time we have xz decompression via `github.com/smira/go-xz` (which is using external process). 

Can we keep one or another, or completely skip `.xz` for now? (until Go support becomes more mature?)

deb/index_files.go
```diff
@@ -23,8 +23,11 @@ type indexFiles struct {
 type indexFile struct {
 	parent       *indexFiles
 	discardable  bool
-	compressable bool
-	onlyGzip     bool
+	ignoreFlat   bool
+	compressOnly bool
```
smira: what's the difference between `ignoreFlat` and `compressOnly` ?

deb/index_files.go
```diff
@@ -95,6 +104,10 @@ func (file *indexFile) Finalize(signer utils.Signer) error {
 	}
 
 	for _, ext := range exts {
+		if file.compressOnly && !(ext == ".gz" || ext == ".bz2" || ext == ".xz") {
```
smira: if `ignoreFlat` removes empty `ext`, what does this check do?

deb/publish.go
```diff
@@ -452,6 +454,36 @@ func (p *PublishedRepo) GetLabel() string {
 	return p.Label
 }
 
+// GetAppStream returns full path list of appstream files
+func (p *PublishedRepo) GetAppStream(component string) ([]string, error) {
+	if utils.Config.AppStreamDir == "" {
```
smira: The way `aptly` is implemented, it is supposed that `Config` object is not accessed from this level.

I would suggest two changes:

* push `AppStreamDir` to `Context`
* pass `AppStreamDir` to this method as parameter (same below in `publish.go`).

btkostner: The `Publish` function below will call the `GetAddonFiles` method. Should I pass the `AppStreamDir` all the way down from the API and CLI calls?

btkostner:
Closing in favor of #473

PR #434, opened 2016-08-11
Add support for flat repositories with no Release file.

ffrank:
In flat repositories, the Release file is supported, but not mandatory,
according to the specification here:

https://wiki.debian.org/RepositoryFormat#Flat_Repository_Format

An example for such a repository is the MapR ecosystem:

http://archive.mapr.com/releases/ecosystem-5.x/ubuntu/binary/

Before this patch, it was not possible to create a mirror of that
repository. The new behavior is to ignore the missing metadata, and
to implicitly disable checksum checks, since there are no checksums to check.

### DIFF
--- a/cmd/mirror_update.go
+++ b/cmd/mirror_update.go
@@ -53,6 +53,11 @@ func aptlyMirrorUpdate(cmd *commander.Command, args []string) error {
 		return fmt.Errorf("unable to update: %s", err)
 	}
 
+	if len(repo.ReleaseFiles) == 0 {
+		context.Progress().Printf("Repository %s has no Release file, disabling checksum verification.\n", repo.Name)
+		ignoreMismatch = true
+	}
+
 	context.Progress().Printf("Downloading & parsing package files...\n")
 	err = repo.DownloadPackageIndexes(context.Progress(), context.Downloader(), context.CollectionFactory(), ignoreMismatch, maxTries)
 	if err != nil {
--- a/deb/remote.go
+++ b/deb/remote.go
@@ -261,6 +261,10 @@ func (repo *RemoteRepo) Fetch(d aptly.Downloader, verifier pgp.Verifier) error {
 		// 0. Just download release file to temporary URL
 		release, err = http.DownloadTemp(gocontext.TODO(), d, repo.ReleaseURL("Release").String())
 		if err != nil {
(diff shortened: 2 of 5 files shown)


jschwartz79:
Do you know if this change will allow aptly to work with http://archive.mapr.com/releases/ecosystem-5.x/ubuntu/ ?  

Also any plans to merge this soon?

smira:
@jschwartz79 I didn't have a chance to look into that yet, but http://archive.mapr.com/releases/ecosystem-5.x/ubuntu/ looks more like regular repo. Just try putting literally apt-sources.list line to `aptly mirror create xxx` and see where it goes ;)

apachelogger:
@ffrank are you still interested in working on this?

From a casual glance, this would benefit greatly from either a system or a unit test.

ffrank:
Sure, I can see about adding some tests. Thanks for picking this up.

From experience, unpatched aptly will not play well with the MapR repositories (or, at all).

ffrank:
After rebasing, the patched code doesn't compile anymore (see Travis output). I can't seem to be able to make any sense of that error. Help?

smira:
@ffrank this identifier was renamed to `http.Error` (https://github.com/aptly-dev/aptly/blob/master/http/http.go#L9)

ffrank:
Thank you @smira. Missed the forest for all the trees.

Looks much better know. LMK if you feel those tests are inadequate or need refinement.

ffrank:
Is the code coverage check a blocking issue?

smira:
Code coverage decrease shouldn't be blocking for us, but I wonder if this new behavior is good - the test you modified for `suslik` is actually hitting a distribution which doesn't exist at all.  So it's not expected that `aptly mirror create` should succeed. I think we might at least try to `HEAD` `Packages` file to make sure it's there?

And probably we need some system test with flat repository which does `mirror update` to make sure there's no regression for flat repositories in the future.

ffrank:
Very good points. Not quite sure how that test should look right now, but it will probably become clear once I sit down and try.

As for checking the Packages file, I looked for an available function for making a HTTP HEAD call from the `http` package, but didn't find one. Should I just make one? (Or just use stdlib from `Fetch`, ignoring the downloader object?)

smira:
@ffrank probably better adding HEAD method to the downloader, it shouldn't be that complicated as existing `Download` methods

Issue #415, opened 2016-06-16
Aptly publish: fork/exec /usr/bin/bzip2: cannot allocate memory


karras:
I'm running an aptly (0.9.7) publish command on a VM (RHEL 6.6) with 4GiB of memory:

`aptly publish -origin=Ubuntu -component=main,restricted,universe,multiverse -distribution=trusty-updates -architectures=amd64,i386,source snapshot trusty-updates-main-stable-20160611T0000Z trusty-updates-restricted-stable-20160611T0000Z trusty-updates-universe-stable-20160611T0000Z trusty-updates-multiverse-stable-20160611T0000Z ubuntu/next-stable`

This will result in the following error:

```
Loading packages...
Generating metadata files and linking package files...
447 / 19076 [====>------------------------------] 2.34 % 2m47
Finalizing metadata files...
ERROR: unable to publish: unable to compress index file: fork/exec /usr/bin/bzip2: cannot allocate memory
```

The aptly process fills more or less the whole memory but doesn't start swapping, beside having 2GiB of swap available.

Why does it not start swapping? Could this be some memory leak / bug?

Thanks!

karras:
The issue doesn't occur with aptly 0.9.5. Seems like this was introduced in 0.9.6 because that version doesn't work either.

Maybe related to #338...

demasiadovivo:
I'm having the same issue after upgrading to 0.9.7. The same repository was working fine when using 0.9.5, but it's failing to publish now. After a couple of seconds aptly consumes all available memory until it crash with the message "ERROR: unable to publish: unable to compress index file: fork/exec /bin/bzip2: cannot allocate memory".
I'm running aptly on ubuntu trusty in a dedicated machine with 2GB of RAM.

erickeller:
I also had a look at the code responsible for generating bz2 contents file seems the following commits are in the suspect list:

```
cave@[~/workspace/go/src/github.com/smira/aptly]$ git tag
 --contains=138f9f7994c2b6fc1470ddd837d709d864a03216
 v0.9.6
 v0.9.7

cave@[~/workspace/go/src/github.com/smira/aptly]$ git tag  --contains=38f4fc209b8ffa6256d475dd2cfaa0c068ae0254
 v0.9.6
 v0.9.7
```

Best Regards

sliverc:
It seems when running publish with flag -skip-contents=true the memory leak doesn't occur

sliverc:
[...]

PR #449, opened 2016-10-14
only create bzip file if needed. #415

sliverc:
Creating bzip of contents files is not necessary but might use up the last memory available.


smira:
👍  thanks!

PR #454, opened 2016-11-07
New contents algorithm with lower memory footprint

sliverc:
Using leveldb to persist contents file paths on file system instead of keeping it in memory.

This way memory footprint decreases significantly. Trade-off a slight slower contents generation especially with large repositories such as Ubuntu.

This fixes issue #415


smira:
Thanks, I think that's the major pain point for users coming to new version of aptly with contents generation

smira:
@sliverc would you mind taking a look at #490, it's almost the same as your implementation, just a bit different in details?

smira:
Merged as #490

PR #458, opened 2016-11-15
Make Travis CI be able to run on forks

sliverc:
When developing a feature it often happens during development that I do not want to create a pull request straight away but do some testing in my fork and when successful I will create a PR.

Unfortunately currently travis doesn't run on a fork due to invalid paths. This PR fixes this so a Fork can run the test unit as well.


smira:
👍

PR #461, opened 2016-11-24
Add systemd service for aptly http server

jacksgt:
Provides a systemd service file for `aptly serve` command.
This file can be placed in `/etc/systemd/system/` (for users/administrators) or `/lib/systemd/system/` (for distributions).
Extra configuration options should be specified in a `.conf` inside `/etc/systemd/systemd/aptly.service.d/`

### DIFF
--- a/aptly-api.service
+++ b/aptly-api.service
@@ -1,0 +1,12 @@
+[Unit]
+Description=APT repository API
+After=network.target
+Documentation=man:aptly(1)
+Documentation=https://www.aptly.info/doc/api/
+
+[Service]
+Type=simple
+ExecStart=/usr/bin/aptly serve api -no-lock -listen=127.0.0.1:8081
+
+[Install]
+WantedBy=multi-user.target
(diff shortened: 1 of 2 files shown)


smira:
Looks good, thank you!

Should we add to Debian package? Or add some description to docs/man page?

rtennill:
I like the idea of providing this service file but I don't think that it should be installed as a service by default. 

The service file should be made available in /usr/share/aptly for users to copy the file, make modifications to suit their needs, and then install or symlink it into the systemd services.

smira:
Ok, so we should probably have two systemd files, one for `aptly serve`, and another for `aptly api serve` ?

I would prefer to keep them binding to localhost by default, as there's no auth in aptly

rtennill:
Yes, there should be one file for each service.

I don't have any experience with the api serve command yet. Are there any conflict to be aware of if both services are installed and running?

jacksgt:
@rtennill Yes, my intention was to provide these as samples, and let distributions/users decide what to do with them.

I'll look into creating a separate service file for `serve` and `api`. I'll also rebase on top of master.

jacksgt:
Awesome, thank you, too!

PR #463, opened 2016-11-28
Added --max-tries option to mirror update command

sliverc:
When downloading several thousands of packages it sometime occurs that a package runs into a CRC error. With a max tries option it retries a single download and can solve such errors in certain situations where the cause is a network issue.

This fixes issue https://github.com/smira/aptly/issues/149

### DIFF
--- a/http/download.go
+++ b/http/download.go
@@ -168,7 +168,7 @@ func (downloader *downloaderImpl) handleTask(task *downloadTask) {
 		}
 	}
 
-	// still an error after retring, giving up
+	// still an error after retrying, giving up
 	if err != nil {
 		task.result <- err
 		return
--- a/http/download.go
+++ b/http/download.go
@@ -52,6 +52,7 @@ type downloadTask struct {
 	result         chan<- error
 	expected       utils.ChecksumInfo
 	ignoreMismatch bool
+	triesLeft      int
 }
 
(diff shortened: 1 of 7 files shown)


http/download.go
```diff
@@ -153,32 +154,59 @@ func (downloader *downloaderImpl) handleTask(task *downloadTask) {
...
 
-	resp, err := downloader.client.Do(req)
+
+	var temppath string
+	for task.triesLeft > 0 {
+
+		temppath, err = downloader.downloadTask(req, task)
+
+		if err != nil {
+			task.triesLeft--
```
smira: I think we should do `os.Remove(temppath)` here in case of error

http/download.go
```diff
@@ -153,32 +154,59 @@ func (downloader *downloaderImpl) handleTask(task *downloadTask) {
...
+		} else {
+			// successful download
+			break
+		}
+	}
+
+	// still an error after retring, giving up
```
smira: typo: `retring`?

http/download.go
sliverc: downloadTask func already takes care of this as only this function knows when the file has actually been created. temppath is always empty if there is an error.

smira: yep, I missed that looking at `os.Rename` below

PR #465, opened 2016-12-08
Allow comment and defaults to be empty when editing a repo through the API

SHyx0rmZ:
As those fields are not required, it should be possible to set them to be empty.


smira:
The idea here was to do partial update, so if you don't set those fields, old values are kept.

With `string` field we can't distinguish empty from unset, but if we change binding structure type to `*string`, we can check for `nil` instead of empty string, so that if field is `nil`, it is unset and we skip updating it, otherwise we do `repo.Comment = *b.Comment`.

smira:
👍  thank you!

SHyx0rmZ:
Thank you, too!

PR #475, opened 2017-01-17
Support a vertical graph layout in addition to the existing horizontal

jola5:
## Requirements
[Build Status](https://travis-ci.org/jola5/aptly)

## Description of the Change
The default layout of graphviz tends to result in very wide graphs for environments with a lot of repositories. Optionally changing graphviz's rankdir results in a more vertical layout which may be preferable in some cases. In using `layout=horizontal` as default this change is fully backwards compatible. 

I decided to use a generic `layout` switch over a simpler `vertical=false` argument. Maybe someone fancies to implement some of graphviz's more sophisticated layouts like radial. 

### horizontal layout (default)
`aptly graph`
`aptly graph -layout=horizontal`
`http://localhost:8080/api/graph.png`
`http://localhost:8080/api/graph.png?layout=horizontal`

### vertical layout (new)
`aptly graph -layout=vertical`
`http://localhost:8080/api/graph.png?layout=vertical`

## Checklist
- [x] documentation updated

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -22,3 +22,4 @@ List of contributors, in chronological order:
 * Phil Frost (https://github.com/bitglue)
 * Benoit Foucher (https://github.com/bentoi)
 * Geoffrey Thomas (https://github.com/geofft)
+* Johannes Layher (https://github.com/jola5)
--- a/api/graph.go
+++ b/api/graph.go
@@ -11,14 +11,20 @@ import (
 	"os/exec"
 )
 
-// GET /api/graph.:ext
+// GET /api/graph.:ext/:vertical
 func apiGraph(c *gin.Context) {
 	var (
 		err    error
 		output []byte
+		vertical bool = false
 	)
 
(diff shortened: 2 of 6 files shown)


smira:
On tests for horizontal vs. vertical layout, as #474 is merged now, can we probably just check raw `.dot` file from endpoint?

jola5:
Yeah sure, definitely sounds like a good idea.

Do you have any input on implementation details? I'm quite unfamiliar with
the whole test architecture. I suppose I would be generating a dot file
from the current test data once manually, then in the test checking the
output of api/graph.dot against the contents of the generated file.

I'll try to update my PR tonight.

jola5:
Fixed the tests. 

But I decided to keep using etree to get the svg image size information. Since this is a pure functional test I really shouldn't care about graphviz's dot file code as long as the horizontal layout is wider than the vertical.  I would have used the png image dimensions in the first place, but using svg is much simpler.

Issue #479, opened 2017-01-19
list deb files


philtay:
Is there any builtin way to list all .deb files (full absolute paths) in a given repository?
To be more clear and as a mere example: I'm looking for the kind of list that can be piped to xargs.
Please consider this as a feature request if the answer to the above question is no.

smira:
Good question, I would imagine we should expose it via https://www.aptly.info/doc/feature/package-display/ interface? So that packages would be part of the context available to the template. Thoughts?

philtay:
Yes, it would be a good solution. Please also consider adding a "sugar" package query to just mean "all packages". Something like `aptly repo search -format="{{.Path}}" myrepo *`.

smira:
It's already there, but not obvious, easiest "always true" filter is `Name` (which means non-empty package name)

philtay:
Ah, ok. Maybe we could support something more obvious, like the lack of a package query means "all". `aptly repo search -format="{{.Path}}" myrepo` instead of `aptly repo search -format="{{.Path}}" myrepo Name`.

smira:
Good idea, thanks!

PR #488, opened 2017-02-10
Allow filter to be empty for `aptly * search` commands

smira:
Empty filter implies "select all packages".

Relates to #479 

## Description of the Change

For family of `aptly * search` commands, allow filter argument to be omitted, so that all the packages are selected.

Special `PackageQuery` is added to select all the package.

## Checklist

- [x] documentation updated


smira:
Doc update: https://github.com/aptly-dev/aptly-dev.github.io/pull/32

PR #490, opened 2017-02-13
Use temporary LevelDB to store contents index

smira:
Attempt to re-implement #454 in a bit different way.

In order to reduce stress on main aptly DB, create temporary levelDB to be used only while generating contents indexes.

Idea and code borrowed shamelessly from #454.

## Description of the Change

Current implementation might use a lot of memory to store map of paths to packages.

## Checklist

- [ ] documentation updated

### DIFF
--- a/deb/contents.go
+++ b/deb/contents.go
@@ -50,7 +50,7 @@ func (index *ContentsIndex) Empty() bool {
 // WriteTo dumps sorted mapping of files to qualified package names
 func (index *ContentsIndex) WriteTo(w io.Writer) (int64, error) {
 	// For performance reasons push method wrote on key per path and package
-	// in this method we now need to merge all pkg with have the same path
+	// in this method we now need to merge all packages which have the same path
 	// and write it to contents index file
 
 	var n int64
--- a/deb/contents.go
+++ b/deb/contents.go
@@ -1,53 +1,59 @@
 package deb
 
 import (
+	"bytes"
+	"errors"
 	"fmt"
-	"github.com/smira/aptly/aptly"
-	"github.com/smira/aptly/utils"
 	"io"
-	"sort"
-	"strings"
+
+	"github.com/smira/aptly/aptly"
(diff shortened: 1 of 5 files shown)


smira:
I did rough test for `xenial` component `main` (only `amd64` arch) publish:

```
       27.43 real        21.18 user         2.92 sys
  95551488  maximum resident set size
         0  average shared memory size
         0  average unshared data size
         0  average unshared stack size
     45540  page reclaims
       260  page faults
         0  swaps
         8  block input operations
       103  block output operations
         0  messages sent
         0  messages received
         6  signals received
      4683  voluntary context switches
     25465  involuntary context switches
```

That's with contents per package pre-cached already, only indexes are generated

smira:
Uncompressed `Contents` index is 63MB.

deb/contents.go
```diff
@@ -1,75 +1,135 @@
...
 func (index *ContentsIndex) Empty() bool {
-	return len(index.index) == 0
+	return !index.db.HasPrefix(index.prefix)
 }
 
 // WriteTo dumps sorted mapping of files to qualified package names
 func (index *ContentsIndex) WriteTo(w io.Writer) (int64, error) {
-	var n int64
+	// For performance reasons push method wrote on key per path and package
+	// in this method we now need to merge all pkg with have the same path
```
sliverc: small typo should be:
all pkgs which have the same path

smira:
thanks @sliverc!

I've added compaction throttling, but only to temporary DB which is being used to build contents index

sliverc:
@smira Great... this looks good!

smira:
Thanks @sliverc!

odyssey4me:
After hitting this issue, I'm wondering when we can expect to see the next release which includes this fix?

smira:
You can try nightly builds (https://www.aptly.info/download/) which already include that fix. I don't have any exact date for the next release so far, but it looks like we're getting close to it

PR #493, opened 2017-02-22
support serving the API over unix domain socket

apachelogger:
`unix://$PATH` as listen argument will bind aptly to a unix domain socket
rather than TCP.

This allows binding the API to a UDS rather than a port.
Since aptly has no concept of authentication or any amount of high level
API hardening one needs to bottle it up in some other manner. Binding
to a localhost port is often a step in the right direction, ultimately is
still a scary insecure setup as any user on that host getting compromised
would mean that the entire archive is compromised as well.
UDS on the other hand are basically files and have their access managed
by regular file permission. As such, binding to a socket is in fact
the least insecure way to listen as you'd have to explicitly open up the
socket permissions to an access qualified group. In the most conservative
scenario that means no one but the aptly user can talk to the API, in a
more practical setup apache might get access as well and proxy the UDS
with authentication or limited to GET operations.

Using UDS allows reducing the attack surface of the API server while
preserving all the flexibility.

Fixes #

## Requirements

## Checklist

- [x] documentation updated

### DIFF
--- a/cmd/api_serve.go
+++ b/cmd/api_serve.go
@@ -2,11 +2,15 @@ package cmd
 
 import (
 	"fmt"
+	"net"
+	"net/http"
+	"net/url"
+	"os"
+
 	"github.com/smira/aptly/api"
 	"github.com/smira/aptly/utils"
 	"github.com/smira/commander"
 	"github.com/smira/flag"
-	"net/http"
 )
 
 func aptlyAPIServe(cmd *commander.Command, args []string) error {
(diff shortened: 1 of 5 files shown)


cmd/api_serve.go
```diff
@@ -34,6 +38,21 @@ func aptlyAPIServe(cmd *commander.Command, args []string) error {
...
+		os.Remove(file)
+		listener, err := net.Listen("unix", file)
+		if err != nil {
+			return fmt.Errorf("failed to listen on: %s\n%s", file, err)
+		}
+		defer listener.Close()
+		err = http.Serve(listener, api.Router(context))
+		if err != nil {
+			return fmt.Errorf("unable to serve: %s", err)
+		}
```
smira: I think there should be `return nil` here, otherwise aptly might fall into regular HTTP serve

system/t12_api/unix_socket.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def run(self):
+        pass
+
+    """
+    Verify we can listen on a unix domain socket.
+    """
+    def check(self):
+        session = requests_unixsocket.Session()
+        r = session.get('http+unix://%s' % urllib.quote(UnixSocketAPITest.socket_path, safe=''))
```
smira: could we  pull here `/api/version` API which always exists and returns HTTP 200 back with simple JSON body?

I think it might be a bit better check to verify that things work correctly

apachelogger:
I've revised the commit to return from socket serving (good catch!) and test the `/api/version` content instead :)

smira:
Code looks good, thanks 👍 

Would you mind adding note on unix domain socket support to `aptly serve` command line help (`make man` should re-generate manual page for you).  And PR for the documentation (website): https://github.com/aptly-dev/aptly-dev.github.io

apachelogger:
I've add a help and manpage update to the commit.

FWIW, the manpage generator doesn't seem to like multi-line `Example:` blocks and loses the newlines. While that looks silly I've kept it this way as the example should be helpful even if incorrectly formatted ^^

apachelogger:
https://github.com/aptly-dev/aptly-dev.github.io/pull/33 for website update

PR #494, opened 2017-02-22
fixes #331, include -dbgsym packages

### LINKED ISSUE #331: no support for -dbgsym packages
In stretch, debian added support for dbgsym packages [1].  Here is some sample output.  Packages are built with debuild.

```
aptly repo include -no-remove-files=true -config=../aptly.conf -repo=stretch artifacts
gpgv: Signature made Wed 06 Jan 2016 10:11:54 PM CST using DSA key ID A8AC8032
gpgv: Good signature from "Gary Kramlich (grim) <grim@reaperworld.com>"
Loading repository stretch for changes file gplugin_0.0.21_amd64.changes...
[+] gir1.2-gplugin-0.0_0.0.21_amd64 added
[+] gplugin_0.0.21_source added
[+] libgplugin-dev_0.0.21_amd64 added
[!] libgplugin-gtk-bin-dbgsym_0.0.21_amd64 has been ignored as it doesn't match restriction
[+] libgplugin-gtk-bin_0.0.21_amd64 added
[+] libgplugin-gtk-dev_0.0.21_amd64 added
[+] libgplugin-gtk0-dbg_0.0.21_amd64 added
[+] libgplugin-gtk0_0.0.21_amd64 added
[+] libgplugin-loaders_0.0.21_all added
[!] libgplugin-lua-dbgsym_0.0.21_amd64 has been ignored as it doesn't match restriction
[+] libgplugin-lua_0.0.21_amd64 added
[!] libgplugin-python-dbgsym_0.0.21_amd64 has been ignored as it doesn't match restriction
[+] libgplugin-python_0.0.21_amd64 added
[+] libgplugin0-dbg_0.0.21_amd64 added
[+] libgplugin0_0.0.21_amd64 added
[!] Some files were skipped due to errors:
  artifacts/libgplugin-gtk-bin-dbgsym_0.0.21_amd64.deb
  artifacts/libgplugin-lua-dbgsym_0.0.21_amd64.deb
  artifacts/libgplugin-python-dbgsym_0.0.21_amd64.deb
ERROR: some files failed to be added
```

I've been trying to figure out how the package restrictions work, but so far haven't been able to grok it enough to fix it.

[1] https://wiki.debian.org/AutomaticDebugPackages

sobczyk:
dbgsym are the new automatic debug packages.
This is an extremely simple change to let these packages in

### DIFF
--- a/deb/changes.go
+++ b/deb/changes.go
@@ -189,6 +189,9 @@ func (c *Changes) PackageQuery() (PackageQuery, error) {
 		binaryQuery = &AndQuery{
 			L: &NotQuery{Q: &FieldQuery{Field: "$PackageType", Relation: VersionEqual, Value: "source"}},
 			R: binaryQuery}
+		binaryQuery = &OrQuery{
+			L: &FieldQuery{Field: "Name", Relation: VersionPatternMatch, Value: "*-dbgsym"},
+			R: binaryQuery}
 	}
 
 	var nameQuery PackageQuery


deb/changes.go
```diff
@@ -189,6 +189,9 @@ func (c *Changes) PackageQuery() (PackageQuery, error) {
 		binaryQuery = &AndQuery{
 			L: &NotQuery{Q: &FieldQuery{Field: "$PackageType", Relation: VersionEqual, Value: "source"}},
 			R: binaryQuery}
+		binaryQuery = &OrQuery{
+			L: &FieldQuery{Field: "Name", Relation: VersionPatternMatch, Value: "*-dbgsym"},
```
smira: I think this breaks `.changes` consistency: aptly verifies that `.changes` is valid and signed by known key, so it tried to avoid pulling packages which are not referenced in `.changes`. This query matches any `*-dbgsym` package.

sobczyk:
ok It actually seems that pkg-create-dbgsym adds it to binary field, i created simple package to test it, and the basic aptly filter works
```
Checksums-Sha1: 
 41eed58a4c4815c3902ffcda1bee4f4d9dbe1b02 5223882 aptly_0.9.7_amd64.deb
 d2288bb923cf0dbcac365b21a3f7c7bc1b8ded0a 1591372 aptly-dbgsym_0.9.7_amd64.ddeb
Checksums-Sha256: 
 9a795048b5033d0a4df08cba9804608df7a0ad14fddb81b8f6fff9339198451b 5223882 aptly_0.9.7_amd64.deb
 05864494200b8f3fb8015b77b389d1ed0d9b5bed7d50f8c7db9c02aca829b4d4 1591372 aptly-dbgsym_0.9.7_amd64.ddeb
Files: 
 0a07649ad9a617bcb63ff841b488980b 5223882 patched/utils optional aptly_0.9.7_amd64.deb
 1c277394fad62459f2480b5d9b7f0ff6 1591372 patched/utils extra aptly-dbgsym_0.9.7_amd64.ddeb
```

and the control:
```
Package: aptly-dbgsym
Source: aptly
Version: 0.9.7
Depends: aptly (= 0.9.7)
Architecture: amd64
Maintainer: xxxx
Installed-Size: 12981
Section: patched/utils
Priority: extra
Description: debug symbols for package aptly
```

smira:
so we don't need this change?

sobczyk:
ok, bad compilation on my side, I'm not an experienced go dev.
I thought binary contains all checksummed blobs.
I'll need to redo the whole code.
I think the best would be to swap Value: c.Binary[0] for Value: c.Files[0].Filename and comparing checksummed filenames.
dbgsym should be only persent when at least one Binary is there.
Another approach is to add separate code blocks for processing ddeb

sobczyk:
Why CollectPackageFiles is called then when there is already a list of acceptable & verified files in the Changes structure?
Thus almost completely negating the need for PackageQuery

smira:
I think the only reason is to do cross-check on package files and contents of `.changes` file. Should we include `-dbgsym` packages based on some other condition which is unique? Like built from the same source package?

sobczyk:
But you copy the files to a tmp dir anyway (Changes.TempDir), so they're the only the files from the changes. There's no point in walking the filesystem 2nd time and then filtering. Maybe only filter out source tar.* packages, as it's in the .dsc. It's pretty much impossible to have different files at that point. (assuming no malicious user swapping them in mid processing, but if that happens the whole repo is compromised anyway)

The only thing I'd change in .changes processing is to add temp dir to aptly config, so it can be on the same filesystem as the rest of the repo. (swap os.TempDir() for some user setting, or default to os.TempDir())
Then copy with reference/hardlink is possible, and verify the checksums after copy to temp dir. That way all files in temp dir could be safely processed.

As a side point , I think I should close this PR, as it's badly done.
Dhould we move this discussion to issue #331?

smira:
Let's move to #331

PR #495, opened 2017-02-24
support systemd activation for `api serve`

apachelogger:
systemd has a feature called socket activation where initially systemd
manages and listens on ports/uds and only invokes a service when traffic
appears. to then hand over the involved sockets, systemd will pass the
relevant FDs into the invoked process and defines them in the environment.

use coreos/go-systemd to grab the active listeners passed by systemd and
use them to serve the api routes. only one listener may be specified right
now as we also only support one -listen argument for the binary.

this allows admins to craft a systemd socket and service file for aptly
where systemd manages the socket, its permission and its live time, and
lazy start aptly when needed.

Fixes #

## Requirements

## Checklist

- [x] documentation updated

### DIFF
--- a/cmd/api_serve.go
+++ b/cmd/api_serve.go
@@ -8,6 +8,7 @@ import (
 	"os"
 
 	"github.com/smira/aptly/api"
+	"github.com/smira/aptly/systemd/activation"
 	"github.com/smira/aptly/utils"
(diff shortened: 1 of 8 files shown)


apachelogger:
Rebased on master and expanded command help + manpage.

Issue #442, opened 2016-09-15
Mirror of repository with only SHA sums produces error


marcusn:
I am trying to mirror the Chef repository:

aptly mirror create -force-components -filter '' chef-trusty-main https://packages.chef.io/stable-apt trusty main

aptly mirror update chef-trusty-main
Downloading https://packages.chef.io/stable-apt/dists/trusty/InRelease...
Downloading https://packages.chef.io/stable-apt/dists/trusty/Release...
Downloading https://packages.chef.io/stable-apt/dists/trusty/Release.gpg...
gpgv: Signature made Wed 14 Sep 2016 06:34:36 PM UTC using DSA key ID 83EF826A
gpgv: Good signature from "CHEF Packages packages@chef.io"
gpgv:                 aka "Opscode Packages packages@opscode.com"
Downloading & parsing package files...
Downloading https://packages.chef.io/stable-apt/dists/trusty/main/binary-amd64/Packages.bz2...
Building download queue...
ERROR: unable to update: unable to compute pool location for filename chef_12.3.0-1_amd64.deb, MD5 is missing

It seems that the Packages.bz2 does not contain MD5sum, but they do contain SHA sums..

camelpunch:
Dupe of https://github.com/smira/aptly/issues/228

smira:
Closing as duplicate.

PR #506, opened 2017-03-19
Configurable pool path hash

### LINKED ISSUE #228: mirrorring depends on non-required MD5sum field of the Binary Package Index
The MD5sum field is a [non-required field](https://wiki.debian.org/RepositoryFormat#Size.2C_MD5sum.2C_SHA1.2C_SHA256) in the Binary Package Index.

Aplty however cannot mirror packages where this field is abscent.  It gives the following error message: "unable to compute pool location for filename 'test.deb', MD5 is missing".

This situation appears for example when you want to mirror a debian repository [hosted on bintray](http://blog.bintray.com/2014/12/16/hosting-debian-packages-on-bintray-rocks/).

tjmerritt:
Fixes #228, #442 

## Requirements

## Description of the Change

This change adds a new configuration field "hashSelector" that defaults to MD5 but can be set to SHA1, SHA256, or SHA512.  The debian package specification does not require the MD5 hash be present, but does require that SHA256 is present.  Aptly currently is using the MD5 hash value from remote repositories to construct path names in the pool directory.  This change allows the hash function to be configured in aptly.conf.  With the change the and the hashSelector configured to SHA256, repositories hosted on Artifactory or BinTray can be mirrored by Aptly.  Without this change they cannot.  

The default behavior of Aptly is unchanged, so there should not be any compatibility issues with existing installations or existing documentation.

## Checklist

- [X] documentation updated

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -26,3 +26,4 @@ List of contributors, in chronological order:
 * Harald Sitter (https://github.com/apachelogger)
 * Johannes Layher (https://github.com/jola5)
 * Charles Hsu (https://github.com/charz)
+* TJ Merritt (https://github.com/tjmerritt)
--- a/aptly/interfaces.go
+++ b/aptly/interfaces.go
@@ -12,15 +12,19 @@ import (
 // PackagePool stores all the package files, deduplicating them.
 type PackagePool interface {
 	// Path returns full path to package file in pool given any name and hash of file contents
-	Path(filename string, hashMD5 string) (string, error)
-	// RelativePath returns path relative to pool's root for package files given MD5 and original filename
-	RelativePath(filename string, hashMD5 string) (string, error)
+	Path(filename string, hash string) (string, error)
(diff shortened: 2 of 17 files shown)


smira:
Thanks, I really like the change. I wonder if there's any reasonable way to have backwards compatibility from MD5 pool? E.g. we could add flag to new packages which are imported with the new hash being used to calculate pool path, and old package would still default to MD5? 

I also thought of changing public pool path to avoid conflicts to something like:

`pool/0a/c5/af34_package_1.3.5.deb` (so that part of hash is prepended to filename, so that even if two packages share same first several octets of the hash and name, version and arch, they can still have different paths in the pool)

smira:
I'm trying to find a way to provide backwards compatibility and reasonable path forward as well. I'm thinking about persisting pool path in the package itself (with fallback to current MD5 path if not provided). That way no matter which hash we pick, we can always find it back in the pool. 

If you don't mind, I would try to use parts of your code.

smira:
Closing this one as it got merged in #539

PR #507, opened 2017-03-20
Added optional usage of symlinks instead of hardlinks when publishing.

seeraven:
Performance of the publish step of aptly was poor when using aptly with large repositories and a NFS-mounted NAS system. After some tests, I discovered that creating hardlinks on the NAS is much slower than using symlinks: For ~40000 links the difference is 18 minutes (hardlinks) vs 2 minutes (symlinks).

## Requirements

## Description of the Change

An option was introduced into the config-file as well as on the command-line (-use-symlinks) to instruct aptly on a per publish base whether to use symlinks or stay with hardlinks. The default option is to use hardlinks.

For the functional testing, I had to add two new check methods (check_file_is_regular and check_file_is_symlink). In addition, I had to modify the tests that compare json results. Apparently it depends on whether python2 or python3 is used, but travis always complained about different results as the json output contained unicode strings, whereas the provided map was UTF-8. By specifying the result directly as json and parsing it using the json.loads() function I was able to get identical results.

## Checklist

- [x] bash completion updated (if applicable) _[pull request will be issued after this one]_
- [x] documentation updated _[inline documentation updated where possible]_


seeraven:
Will restructure fork and open again.

PR #519, opened 2017-03-23
Feature: Option to publish using symlinks instead of hardlinks

seeraven:
Performance of the publish step of aptly was poor when using aptly with large repositories and a NFS-mounted NAS system. After some tests, I discovered that creating hardlinks on the NAS is much slower than using symlinks: For ~40000 links the difference is 18 minutes (hardlinks) vs 2 minutes (symlinks).

## Requirements

## Description of the Change

An option was introduced into the config-file as well as on the command-line (-use-symlinks) to instruct aptly on a per publish base whether to use symlinks or stay with hardlinks. The default option is to use hardlinks.

For the functional testing, I had to add two new check methods (check_file_is_regular and check_file_is_symlink). In addition, I had to modify the tests that compare json results. Apparently it depends on whether python2 or python3 is used, but travis always complained about different results as the json output contained unicode strings, whereas the provided map was UTF-8. By specifying the result directly as json and parsing it using the json.loads() function I was able to get identical results.

## Checklist

- [x] documentation updated


smira:
First of all, thanks for taking time to implement feature, it's one of the most requested ones.

I was thinking of providing same feature on a bit different level - aptly supports multiple publish roots: one is default (filesystem with hardlinks), multiple S3, Swift endpoints. I thought that right way to support symlinks/hardlinks and multiple publish directories was to open ability to have multiple filesystem endpoints, each with its own root and options to use either hardlinks or softlinks or even simply doing file copy.

What do you think of that?

seeraven:
I like that approach. It would introduce a lot more flexibility and fit in nicely with the other endpoints.

So after a quick look, the main modification would be to extend the files published storage provider constructor to take an extra argument about what kind of link action it should use, and provide the infrastructure via the config. And of course the tests. :-)

I'll look into it and start on a new branch on my fork.

seeraven:
Closed as new pull request is there with discussed implementation: https://github.com/smira/aptly/pull/521 .

PR #520, opened 2017-03-24
Add option -skip-existing-packages to mirror update to speed up download queue setup

seeraven:
When updating a mirror, all package files are downloaded and a list of downloads is generated by checking which files are already on disc and have the right size. This can take some time, especially on storage systems like NAS or external hard drives when mirroring large repositories.

## Requirements

## Description of the Change

To speed up the download queue setup, the internal database for the mirror can be used. It is only updated after a successful mirror update and contains the list of packages of the mirror. Assuming nobody interfered and deleted files manually, we can use this database to check what packages need to be downloaded. This new functionality is enabled only when using the command line option -skip-existing-packages, otherwise the behavior is unchanged.

## Checklist

- [x] documentation updated


man/aptly.1
```diff
@@ -1,7 +1,7 @@
 .\" generated with Ronn/v0.7.3
 .\" http://github.com/rtomayko/ronn/tree/0.7.3
 .
-.TH "APTLY" "1" "March 2017" "" ""
+.TH "APTLY" "1" "2017-03-23" "" ""
```
smira: this seems to be generated with _broken_ ronn, `make man` should install correct `ronn` fork for you and run man page generation

seeraven: Took a while to get ronn up and running. Apparently using mustache in version 1.0.4 caused errors, so I had to revert that to 1.0.3. Nevertheless, it now seems to work and I've fixed the man page in 4ddf85b.

seeraven:
I've opened the pull request https://github.com/aptly-dev/aptly-dev.github.io/pull/38 for the web docs.

smira:
`aptly_1.0.0+8+g47b5cc2_i386` nightly got released which includes this feature

Issue #173, opened 2015-01-19
Allow using soft links for publishing


qwach:
Aptly currently uses hard links to publish files from local repositories. Unfortunately, hard links aren't always possible - for example, OpenAFS doesn't allow hard links between files in different directories. Please add an option to use soft links instead.

qwach:
This is mostly #120. Just fall back to soft links if hard links don't work, then to plain copying if soft links don't either?

smira:
Probably it should be configuration option.

d9pouces:
Is this enhancement still planned for a future version?

dpedu2:
+1

PR #510, opened 2017-03-22
Allow published repo to reside on different filesystem

l1k:
## Requirements

All new code should be covered with tests, documentation should be updated. CI should pass.

## Description of the Change

Currently packages are hardlinked from the internal pool to the published repo, which requires them to reside on the same filesystem. This precludes mounting the published repo from a remote machine, e.g. via NFS or sshfs, which would be desirable to physically separate the machine hosting the internal pool and signing key from the machine serving the published repo.

Autodetect such a setup and, if found, copy files instead of using hardlinks.

Add helpers to check whether two given paths reside on the same filesystem and to check whether an already published package file is identical to the one to be published by comparing their MD5 hashes.

## Checklist

- [ ] unit-test added (if change is algorithm)
- [ ] functional test added/updated (if change is functional)
- [ ] man page updated (if applicable)
- [ ] bash completion updated (if applicable)
- [ ] documentation updated
- [ ] author name in `AUTHORS`

PR #521, opened 2017-03-24
Implemented local filesystem endpoint with support for hardlinks, symlinks and copy

### LINKED ISSUE #120: Publications to different filesystems
I'd like to have an ability to publish my sanapshots onto different filesystems. Now it's impossible because aptly try to manage hard links, but in case of another file system, it shall use **copy** command.

seeraven:
Rewritten the previous pull request to support multiple local filesystem endpoints, each with a customizable root directory and a link method (hardlinks, symlinks or file copies).

## Requirements

## Description of the Change

The local filesystem endpoints are configured in the configuration file similar to the s3 or swift endpoints. For example, a snippet like the following adds three endpoints, each with a different method to "link" files from the pool to the public directory:
```
"LocalPublishEndpoints": {
               "test1": {
                 "rootDir": "/opt/srv1/aptly_public",
                 "linkMethod": "symlink"
               },
               "test2": {
                 "rootDir": "/opt/srv2/aptly_public",
                 "linkMethod": "copy"
               },
               "test3": {
                 "rootDir": "/opt/srv3/aptly_public",
                 "linkMethod": "hardlink"
               }
             }
```
Backwards-compatibility is fully preserved, so without specification of the endpoint the default hardlinked public directory is used.

To specify the new endpoint, the prefix _local_ is used, e.g.,
`aptly publish snapshot wheezy-main local:test1:wheezy/daily`

Fixes #120, #173

## Checklist

- [x] documentation updated


files/public.go
```diff
@@ -104,12 +125,19 @@ func (storage *PublishedStorage) LinkFromPool(publishedDirectory string, sourceP
...
 
-		srcSys := srcStat.Sys().(*syscall.Stat_t)
-		dstSys := dstStat.Sys().(*syscall.Stat_t)
-
-		// source and destination inodes match, no need to link
-		if srcSys.Ino == dstSys.Ino {
-			return nil
+		if storage.linkMethod == LinkMethodCopy {
+			// if source and destination have the same size, no need to copy
+			if srcStat.Size() == dstStat.Size() {
```
l1k: Seems risky to only rely on size. In https://github.com/l1k/aptly/commit/72dd4ba0253728832338485b441f815f08edc63c#diff-6e0cc192367b696b238356fae51fdc00R122 I compute the MD5 of the destination but that's obviously expensive. This should probaby be configurable, I could do that in a separate commit.

man/aptly.1.ronn.tmpl
```diff
@@ -121,6 +135,25 @@ Options:
...
+and the following associated settings:
+
+   * `rootDir`:
+     The publish directory, e.g., `/opt/srv/aptly_public`.
+   * `linkMethod`:
+     This is one of `hardlink`, `symlink` or `copy`. It specifies how aptly links the
+     files from the internal pool to the published directory.
+     If not specified, empty or wrong, this defaults to `hardlink`.
+
+In order to publish to such a local enpoint, specify the endpoint as `local:endpoint-name`
```
l1k: s/enpoint/endpoint/

l1k:
From a quick review this mostly LGTM. One nit I have is that "local" can be misleading as the filesystem can be mounted from a remote location, so a more generic term like "filesystem" might be more apt. Again, this is just a nit.

Unfortunately I won't be able to test this before Monday. :( You're way too fast for me. ;)

man/aptly.1.ronn.tmpl
seeraven: Good catch! Fixed in 027e865

files/public.go
seeraven: You're right, the only reason I've used the size instead of a checksum was performance. Adding it optionally is a good idea, I'll try that tomorrow. Perhaps the source md5 is already in the database and we only have to compute the md5 of the destination.

l1k: The sourceMD5 is already passed in to this function.

smira:
Thanks for your contribution! 

I'm going to push out new aptly release 1.0.0, most probably it would take today and tomorrow to finish that. I will get back to this PR as soon as I'm done with the release work.

l1k:
I've just tested both the copy and the symlink functionality. Seems to work like a charm.

files/public.go
```diff
@@ -104,12 +125,19 @@ func (storage *PublishedStorage) LinkFromPool(publishedDirectory string, sourceP
...
+				return nil
+			}
+		} else {
+			srcSys := srcStat.Sys().(*syscall.Stat_t)
+			dstSys := dstStat.Sys().(*syscall.Stat_t)
+
+			// if source and destination inodes match, no need to link
+			if srcSys.Ino == dstSys.Ino {
```
smira: With hardlinks, it's only possible to create them across the the same filesystem, with symlink, it's no longer true

I don't remember off top of my head if there is good and easy way to verify it's the same filesystem exactly, but we could at lest check that `.Dev` fields match

l1k: It seems that the symlink is followed and thus `dstSys` is the `stat_t` of the source file. At least that's what I observed when I tested this. Hence the code seems to be correct and is actually quite elegant. But I'm not 100% sure that I'm recalling this correctly, so it would be good if someone else could double-check.

One issue I did observe was that for hardlinks, there's no upfront testing if the published directory is on the same filesystem as the internal pool. If users don't take care of this, they'll only get an error when they're in the middle of publishing the repo. It might be worth to check this earlier. Code to check if two paths are on the filesystem is part of my pull #510. I could rework this and send a pull to @seeraven's branch or he could cherry-pick the commit onto his branch and throw out the unneeded bits.

smira: Right, symlink is followed, but if it's symlink  it might point to file on _different_ filesystem with _same_ inode. So that check wouldn't make sense if filesystem for source/destination aren't same.

l1k: I see, good point. Indeed taking the `.Dev` fields also into account should address this concern.

seeraven:
So, finally I've got some time to work further on this. First, I'll implement the md5 checksum check for the copy method with an option to disable it explicitly.
Then the question remains how to call the endpoint. As @l1k said correctly, _local_ seems to be quite general. I've chosen the name mainly because it was shorter than _filesystem_ or something like that. Any suggestions?

smira:
`filesystem` sounds better than `local` for me, as aptly already has `local repo` vs. `mirror`.

seeraven:
Implemented the md5 check for the copy method and added an option to switch back to a size check if the user explicitly enables it (config file option `verifyMethod`). I also added a test to ensure the md5 variant determines changes for files at the same size where the `size` variant fails.
The endpoint is now called `FileSystemPublishEndpoint` internally and on the command line the prefix is `filesystem`, e.g., `filesystem:nfsshare`.

smira:
Another idea I had about naming (I'm really bad at naming things) was `mountpoint` or something along those lines, so that you could mount different filesystem as publish roots.

smira:
And if it's good for you to merge this PR, could you please squash commits a bit to keep `master` history clean? Thanks.

seeraven:
Sure, no problem. Glad to have it accepted. I will merge with master first and rebase afterwards. ;-)

seeraven:
I've squashed it into one commit, so its ready to merge.

PR #539, opened 2017-04-04
New package pool with configurable hashing base

smira:
Ref: #506 

## Requirements

aptly is relying on MD5 hash in many places:

* placing files in package pool
* checking for changes while uploading to S3, Swift

At the same time MD5 is not required hash for Debian packages. We need to be more flexible in the way we select hashes, at the same time backwards compatibility should be preserved.

Problems:

* downloader is downloading directly to package pool, so it picks up destination filename before download starts
* without MD5 S3/Swift would be useless
* package conflicts with name clashes in package pools

## Description of the Change

- [x] refactor HTTP downloader to be single-threaded
- [x] refactor package pool so that it doesn't expose filenames
- [x] update published repositories to the new package pool
- [x] update package file import code 
- [x] re-implement multi-threaded mirror downloader
- [x] calculate required checksums while downloading and importing
- [x] switch to SHA256 as default checksum
- [x] add progress bar on mirror update package saving
- [x] ChecksumCollection
- [x] fill checksums on ImportFile
- [x] Verify method moved to PackagePool
- [x] skip existing pkgs in mirror update should fill back checksums
- [x] disable legacy path support in fresh aptly installation
- [x] restore ability for concurrent mirror update

## Checklist

- [x] documentation updated


smira:
This is WIP, it doesn't even build now

PR #577, opened 2017-05-23
Added support for NotAutomatic, ButAutomaticUpgrades and Origin fields

### LINKED ISSUE #184: Reuse all field from a mirror Release file in the generated release

sliverc:
As a follow up on PR https://github.com/smira/aptly/pull/363 this PR fixes #184.

## Description of the Change

The discussion in #363 resulted in that additional fields in Release file are fairly specific and are mainly handled by aptly itself exception being NotAutomatic, ButAutomaticUpgrades and Origin fields. What would be desirable though is to inherit those fields from mirror directly.

So what I implemented in this PR is:
* automatically inheriting Origin, NotAutomatic and ButAutomaticUpgrades fields from mirror, in case all snapshots which are being published have the same value of this field.
* if a publishes consist of several snapshots with different values of those fields, in case of NotAutomatic and ButAutomaticUpgrades those won't be added (same behaviour as today) and Origin will be generated with prefix and distribution (same behaviour as today)
* those 3 flags can always be overwritten during publish with command line flags -origin, -notautomatic, -butautomaticupgrades

## Checklist

- [x] documentation updated


sliverc:
Update for documentation: https://github.com/aptly-dev/aptly-dev.github.io/pull/44

dl.bintray.com seems to be a bit flaky today hence some tests failed. I guess best to rerun again at a later point.

PR #608, opened 2017-08-02
Filter between mirrors

### LINKED ISSUE #357: -filter-with-deps across mirrors
I was trying to make a snapshot of selected Debian Jessie packages, getting the latest version of each by mirroring the main Jessie repo plus updates, backports, and security. (Similar to the "mirroring with snapshots" tutorial.)

This exposes a problem when mirroring using -filter-with-deps, snapshotting, and then merging. Wireshark 1.x is in the main Jessie repo, but backports contains Wireshark 2.x. Both repos contain the wireshark-qt package, but the 2.x version has dependencies that the 1.x version does not. Said dependencies are in the main repo, but not the backports repo. So when I merged the snapshots together, 'aptly snapshot verify' failed because the new dependencies aren't in any of the source snapshots that were merged together. Attempting to install Wireshark from the resulting snapshot fails for the same reason.

I don't see any way to -filter-with-deps across multiple mirrors, so I can only see two workarounds:
1. Mirror the entirety of all of the repositories. This is not feasible due to the amount of bandwidth and disk space that would be consumed.
2. Add the missing dependencies to the filter configuration for each mirror. This works, but means you have to remember to verify your snapshots every time you update your mirrors, in order to find out if you're missing any dependencies. Then you need to update your mirror filters, rinse and repeat until verification succeeds.

It would be nice if aptly handled this situation better. Maybe users could define a "meta-mirror" that takes multiple URLs and presents a combined package list, which could then be filtered as a whole. This behavior would be more apt-like, since apt will search all the sources it has available when trying to locate a package.

mswart:
Fixes #357

## Description of the Change

Distributions contain a huge set of packages, mirroring them requires a large amount of storage and bandwidth. To mitigate this issues aptly provides an filter feature to reduce the set of mirrored packages. Especially it supports to limit the downloaded packages to a set of packages and their needed dependencies as well.

As this filter is self-contained within each mirror, it has two limitations:

1. **its difficult to partial mirror a security or updates mirror to match a filtered main repository.** Update and security repositories contain (newer) packages for a (small) subset of the whole distribution to fix bugs/security issues. As the repository contains not all packages, filtering breaks due to cuts within the dependency tree.

2. **integration of packages from other sources.** It is not possible to filter for all dependencies of a package if this package itself is not part of the mirror repository but instead published within another (local) repository.

In both cases all dependencies between repositories/mirrors have to be included manually in the mirror filter.

This commit introduces `-deps-from-mirrors` and `-deps-from-repos` options for mirrors. They are comma-separated lists referencing mirror/repository names.

If `-filter-with-deps` is activated all dependencies of packages from the references mirrors/repositories that are part of the current mirror are included in addition to the filter result itself. This resolves 2.

If in addition `-dep-follow-all-variants` is activated, all packages that are included in the referenced mirrors/repositories and present in the current mirror (matched by package name only) are selected to download, too. This resolves 1.

## Limitations / Questions

- I am not confident about the option names, but I lack ideas for shorter, more to the point ones.
- The implemented approach differs from the proposals in #357 - each mirror is updated on its own. If both mirrors have (many) packages with dependencies only in other mirrors, it might take some update rounds (update mirror 1, update mirror 2, update mirror 1 ...) to pick up all needed dependencies. I observed a similar behavior for experiments with xenial-mirrors and its security/update repositories. It took 2-3 updates to reach a stable filter point. I am currently unsure about the cause, I wound investigate further if the feature / fundamental implementation design is approached.
- I am unsure how to best implement functional tests.

## Checklist

- [ ] documentation updated

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -64,3 +64,4 @@ List of contributors, in chronological order:
 * Golf Hu (https://github.com/hudeng-go)
 * Cookie Fei (https://github.com/wuhuang26)
 * Andrey Loukhnov (https://github.com/aol-nnov)
+* Malte Swart (https://github.com/mswart)
--- a/api/api.go
+++ b/api/api.go
@@ -228,7 +228,7 @@ func showPackages(c *gin.Context, reflist *deb.PackageRefList, collectionFactory
 		list.PrepareIndex()
 
 		list, err = list.Filter([]deb.PackageQuery{q}, withDeps,
-			nil, context.DependencyOptions(), architecturesList)
+			nil, nil, context.DependencyOptions(), architecturesList)
 		if err != nil {
 			AbortWithJSONError(c, 500, fmt.Errorf("unable to search: %s", err))
 			return
--- a/api/api.go
+++ b/api/api.go
@@ -244,7 +244,7 @@ func showPackages(c *gin.Context, reflist *deb.PackageRefList, collectionFactory
 				fmt.Println("filter packages by version, query string parse err: ", err)
 				c.AbortWithError(500, fmt.Errorf("unable to parse %s maximum version query string: %s", p.Name, err))
 			} else {
-				tmpList, err := list.Filter([]deb.PackageQuery{versionQ}, false,
+				tmpList, err := list.Filter([]deb.PackageQuery{versionQ}, false, nil,
 					nil, 0, []string{})
 
 				if err == nil {
--- a/api/mirror.go
+++ b/api/mirror.go
@@ -306,7 +306,7 @@ func apiMirrorsPackages(c *gin.Context) {
 
 		list.PrepareIndex()
 
-		list, err = list.Filter([]deb.PackageQuery{q}, withDeps,
+		list, err = list.Filter([]deb.PackageQuery{q}, withDeps, nil,
 			nil, context.DependencyOptions(), architecturesList)
 		if err != nil {
 			AbortWithJSONError(c, 500, fmt.Errorf("unable to search: %s", err))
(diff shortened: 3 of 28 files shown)


smira:
Thanks, looks interesting!

mswart:
Nice. I have rebased the PR to fix the merge conflicts.
I am happy to make further adjustments or addition to your feedback.

belm0:
I tried the patch with my use case and it worked well-- looking forward to the feature.

It's been a while since the last rebase, would you run again?

erikoest:
Is this patch available for aptly version 1.5.0? I tried to merge the mswart:filter-between-mirrors branch over aptly-dev:v1.5.0 and got a few conflicts, mainly in the api code.

neolynx:
Hi !

It looks like this was never merge, just closed probably..

should we reopen this ?

erikoest:
It's up to the aptly-developers, I guess. And I think a new merge req is needed since this one won't fit with newer versions of aptly. I'm using the filter-between-mirrors functionaliy actively, so I'll need it in later versions. But first, I can ask mswart if he would like to maintain his branch and make it compatible with v1.5.

neolynx:
rebased...

neolynx:
does not seem to build anymore on master

neolynx:
@erikoest 
I read about some issue here: https://github.com/aptly-dev/aptly/issues/357

were they fixed, could they still be an issue ?

erikoest:
I'm not familiar with issue #357. I don't think we've seen this problem in our systems. Maybe you should consult @mswart who wrote the patch.

neolynx:
I managed to fix the build and the tests :)

coverage is low, so we need more tests: 38.52% of diff hit (target 74.84%) 

@mswart I know it has been years, would you still be interested in taking this further ?

neolynx:
rebased on master...

PR #620, opened 2017-08-16
Add -skip-cleanup option for publish commands.

### LINKED ISSUE #570: Ability to skip prefix cleanup on publish operations
I am publishing multiple large snapshot (e.g. Ubuntu security snapshots) under the same prefix.

I have noticed that aptly became very slow when switching and dropping publishes in that prefix, and the slow part is the "Cleaning up prefix "[...]" components main...".

If I am doing multiple switches or drops, that will take a lot of time and would not re really needed.

I am suggesting adding a -skip-cleanup component to the publish command, that would skip cleanup on all publish operations, and to add the ability to invoke the cleanup of a prefix on demand (aptly prefix cleanup).

Thanks!

cavedon:
Fixes #570.

## Description of the Change

When  publishing multiple large snapshot (e.g. Ubuntu security snapshots) under the same prefix aptly becames very slow when switching and dropping publishes in that prefix, and the slow part is the "Cleaning up prefix "[...]" components main...".

In case of it is necessary to update/switch/drop multiple published repos/snapshots, a significant amount of time can be saved by performing the cleanup phase only after the last one operation.

This change allow skipping the cleanup of unreferenced files on publish switch/update/drop via the -skip-cleanup command line option. It also adds support for the API SkipCleanup parameter.

## Checklist

- [ ] documentation updated


smira:
Thanks!

cavedon:
Thank you for considering this change for merge!
I have used the web UI to resolve the conflict on AUTHORS.

PR #621, opened 2017-08-16
Fix the doubling of the s3 prefix

### LINKED ISSUE #619: S3 publishing endpoint appears to reupload everything
I am investigating slowness around publishing new snapshots to s3. I turned on debugging and the ListObjects call appeared to double up on the prefix ie 
`GET /?max-keys=1000&prefix={prefix}%2F{prefix}%2F`. 

As far as I can tell I have the s3 endpoint configured correctly and I am not sending a prefix in the command but do have the prefix configured in the endpoint. 

Unless I am reading the code incorrectly it looks like that is happening as a result of these 2 lines:
https://github.com/smira/aptly/blob/master/s3/public.go#L270
https://github.com/smira/aptly/blob/master/s3/public.go#L325

MMartyn:
Fixes #619

### DIFF
--- a/s3/public.go
+++ b/s3/public.go
@@ -267,7 +267,7 @@ func (storage *PublishedStorage) LinkFromPool(publishedDirectory, baseName strin
 	poolPath := filepath.Join(storage.prefix, relPath)
 
 	if storage.pathCache == nil {
-		paths, md5s, err := storage.internalFilelist(storage.prefix, true)
+		paths, md5s, err := storage.internalFilelist("", true)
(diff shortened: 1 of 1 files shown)


MMartyn:
@smira I wasn't sure how to go about writing a test for this as all of the behavior is private to the method. However, I did confirm that the current tests pass.

smira:
thanks, @MMartyn! I'll take a look if we can add unit-tests for it

smira:
thanks, merged that change in #622

PR #630, opened 2017-09-05
Expose the context outside of the cmd package.

ringods:
## Requirements

## Description of the Change

Skyscrapers wants to contribute back what we use from the community. We are a heavy user of [Concourse](http://concourse.ci/). Concourse supports monitoring changes in different artifacts by way of resource types. We are contributing a new resource type to monitor the release of new versions of a package and trigger builds when that happens:

https://github.com/skyscrapers/apt-package-resource

We are still in development for this resource, but so far I had to make one change to the `aptly` code base to support is in reusing the code: have access to the command context.

This resource type is my first Go contribution which means that also this PR is possibly not the way you should do things in Go. I'm open to learn how to do it better then.

## Checklist

- [ ] documentation updated

### DIFF
--- a/cmd/context.go
+++ b/cmd/context.go
@@ -29,3 +29,8 @@ func InitContext(flags *flag.FlagSet) error {
 
 	return err
 }
+
+// GetContext gives access to the context
+func GetContext() *ctx.AptlyContext {
+	return context
+}


ringods:
@smira rebased. But I see that that last travis build on `master` failed, so the build results from my PR won't be any better. ;-)

https://travis-ci.org/smira/aptly/builds/276999193

smira:
thanks

PR #658, opened 2017-10-31
make deb reader handle new control.tar options introduced in dpkg 1.17.6

### LINKED ISSUE #655: Aply cannot handle deb packages using control.tar.xz
## Detailed Description

Ever since the release of dpkg 1.19.0, `dpkg-deb` enables `--uniform-compression` by default. This means `control.tar*` file will be compressed using `xz` by default (to be more precise, using the same compression method as `data.tar*` file). This claims `control.tar.xz` file inside `.deb` file instead of `control.tar.gz`.

Currently aptly does not seem to support handling `control.tar.xz` inside `.deb` file.

## Context

Using `.deb` packages built using `dpkg (>= 1.19.0)` will encounter this problem.

Currently it affects users of Debian Testing and Debian Unstable but we can expect more users affected by this bug due to downstream adaption of new dpkg.

## Possible Implementation
Do not assume hardcoded `control.tar.gz` inside `.deb` files. Instead, we should do an intelligent analysis about the compression method used for `control.tar.*` and `data.tar.*`.

## Your Environment
Debian Unstable, using dpkg 1.19.0.4.

apachelogger:
newly supported is uncompressed control.tar and xz compressed
control.tar.xz. latter is used by ubuntu for dbgsym ddebs.

Fixes #655

## Requirements

## Description of the Change

Currently Ubuntu 18.04 builds with ddebs (debug symbols) cannot be uploaded to aptly as they get rejected with 

```
    Unable to read file /home/me/.aptly/upload/Aptly__Repository-smith-20171031-18307-yjnsbr/test.ddeb: unable to find control.tar.gz part in package /home/me/.aptly/upload/Aptly__Repository-smith-20171031-18307-yjnsbr/test.ddeb
```

which is caused by the packages making use of control.tar.xz members which are fairly newly supported in dpkg. The proposed change adds support for new control.tar variants as per deb format specification manpage.

## Checklist


apachelogger:
Travis seems to fall over gometalinter, not sure why or what to do about that.

apachelogger:
Linting fixed by #659

apachelogger:
rebased & travis green now

PR #659, opened 2017-10-31
fix linting by using new maligned linter instead of aligncheck

apachelogger:
upstream switched the alignment check backend and in doing so fails to run
if the old backend is defined in the config.

also skip alignment linting on a struct we use for byte decoding as we have
no choice in its member order.

Also see https://github.com/alecthomas/gometalinter/commit/90d987842c8b2b8591691f6227d855a3a992e202

Fixes all PRs failing on linting


smira:
Thanks!

PR #677, opened 2017-11-21
Allow editing of mirror archive url

sliverc:
## Description of the Change

Allow editing of mirror archive url. This is handy in case a mirror is down or slow and user wants to move to another identical mirror without loosing snapshots.

## Checklist

- [x] documentation updated


sliverc:
Added documentation in PR aptly-dev/aptly-dev.github.io#57

system/t04_mirror/EditMirror10Test_gold
```diff
@@ -0,0 +1,2 @@
+Downloading ftp://ftp.ru.debian.org/debian/dists/wheezy/Release...
```
smira: I think it should be downloading here from `ftp.ch.debian.org` here? (New URL?)

sliverc: Indeed that's not correct. Preparation of archiveRoot was missing. Should be fixed now.

PR #683, opened 2017-11-29
Use Go context to abort gracefully mirror updates

### LINKED ISSUE #545: aptly should clean up after itself on interrupts
There are many aptly actions (like publishing or updating mirrors) which create temporary stuff while they're running. aptly should clean up after itself if it gets aborted.

Splitting it from #543 

## Possible Implementation

Good thing would be to use `context` package to include context which gets canceled when aptly process is interrupted (same in API for client walkaways).

smira:
There are two fixes here:

1. Abort package download immediately as ^C is pressed.
2. Import all the already downloaded files into package pool,
so that next time mirror is updated, aptly won't download them
once again.

Fixes #545

## Checklist

- [ ] documentation updated

### DIFF
--- a/context/context.go
+++ b/context/context.go
@@ -32,7 +32,7 @@ import (
 type AptlyContext struct {
 	sync.Mutex
 
-	ctx gocontext.Context
+	gocontext.Context
 
 	flags, globalFlags *flag.FlagSet
 	configLoaded       bool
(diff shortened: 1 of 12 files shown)


context/context.go
```diff
@@ -30,6 +32,8 @@ import (
 type AptlyContext struct {
 	sync.Mutex
 
+	ctx gocontext.Context
```
sliverc: I was just browsing through this change. Thanks for working on this issue.

One question I have is though wouldn't it be better and more readable when go context is an embedded type of Aptly context (like sync.Mutex)? This way aptly and go context can be used interchangeably.

smira: Interesting idea, I never thought about it to be honest. I'll try it, thanks!

PR #696, opened 2018-01-15
properly expose AcquireByHash through the api

apachelogger:
- new publish calls can now enable AcquireByHash by right away (previously
  one would have had to create a new publishing endpoint and then
  explicitly switch it to AcquireByHash)
- all json marshals of PublishedRepo now contain AcquireByHash (allows
  inspecting if a given endpoint has AcquireByHash enabled already; also
  enables verification that a switch/update actually applied a
  potential AcquireByHash change
- update all tests to reflect that default state of AcquireByHash
- update creation and switch testing to explicitly toggle AcquireByHash to
  make sure state mutation works as expected

## Requirements

## Description of the Change

AcquireByHash is rubbish to use via API without this.

## Checklist

- [x] documentation updated (https://github.com/aptly-dev/aptly-dev.github.io/pull/61)


apachelogger:
I am not quite sure why that test fails. it seems fairly unrelated -.-

PR #705, opened 2018-02-19
prevent removal of a PublishedStorage's root dir

apachelogger:
## Requirements

## Description of the Change

presently there is no use case where we need this. on the other hand,
passing empty paths into any of the remove methods is indicative of a bug.
this is particularly dangerous as this can temporarily smash the publish
root but later restore it again when actually publishing. this makes
for super nasty and hard to track down problems.

to guard against this simply disallow root dir removal using empty
strings. should we find a use case for this in the future we can always
revisit this (FTR: I think very explicitly API should be used so everyone
knows what is going on and you can't accidentally run it)

## Checklist

### DIFF
--- a/files/public_test.go
+++ b/files/public_test.go
@@ -320,3 +320,19 @@ func (s *PublishedStorageSuite) TestLinkFromPool(c *C) {
 	err = s.storageCopySize.LinkFromPool(filepath.Join("", "pool", "main", "m/mars-invaders"), "mars-invaders_1.03.deb", pool, srcPoolPath, sourceChecksum, false)
 	c.Check(err, IsNil)
 }
+
+func (s *PublishedStorageSuite) TestRootRemove(c *C) {
+	// Prevent deletion of the root directory by passing empty subpaths.
+
+	pwd := c.MkDir()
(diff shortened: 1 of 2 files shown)


files/public_test.go
```diff
@@ -316,3 +316,19 @@ func (s *PublishedStorageSuite) TestLinkFromPool(c *C) {
...
+
+	// Symlink
+	linkedDir := filepath.Join(pwd, "linkedDir")
+	os.Symlink(s.root, linkedDir)
+	linkStorage := NewPublishedStorage(linkedDir, "", "")
+	c.Assert(func() { linkStorage.Remove("") }, PanicMatches, "trying to remove empty path")
+
+	// Actual dir
+	dirStorage := NewPublishedStorage(pwd, "", "")
+	c.Assert(func() { dirStorage.RemoveDirs("", nil) }, PanicMatches, "trying to remove empty path")
```
smira: panic message doesn't match the code (`trying to remove the root directory`)

apachelogger:
Ah yes indeed.

Rebased and fixed.

PR #715, opened 2018-03-12
Replace defunct keys.gnupg.net references with MIT keyserver.

belm0:


smira:
Looks like pgp.mit.edu is down as well, going to replace with https://sks-keyservers.net/overview-of-pools.php

belm0:
Thank you-- shall I update this PR to pool.sks-keyservers.net?

smira:
I'll do PR which covers that & new aptly signing key, thanks a lot!

Issue #657, opened 2017-10-31
Gpg 2.1 compatiblity


michael-todorovic:
Hello,
I'm on Debian Stretch (gpg 2.1). I generated my gpg key so I have a .gnupg/pubring.kbx instead of .gnupg/pubring.gpg and .gnupg/secring.gpg. The gpg provider doesn't work but I can't get the internal provider to work.
`host:~$ aptly publish repo -batch -gpg-key=<blah> -passphrase-file="/data/aptly/.DoNotRemoveMandatoryForAptly" -gpg-provider=internal <name> <prefix>
opengpg: failure opening keyring '/data/aptly/.gnupg/pubring.gpg': open /data/aptly/.gnupg/pubring.gpg: no such file or directory
opengpg: failure opening keyring '/data/aptly/.gnupg/secring.gpg': open /data/aptly/.gnupg/secring.gpg: no such file or directory
ERROR: unable to initialize GPG signer: couldn't find key for key reference <blah>`

I tried to specify -secret-keyring with the kbx and it didn't work either:
`ERROR: unable to initialize GPG signer: error load secret keyring: openpgp: invalid data: tag byte does not have MSB set`

Am I doing something wrong or does it need a fix (I may submit a PR then)?
Thanks!

smira:
aptly doesn't support gpg 2.1 both when using external `gpg` program & when using internal OpenPGP implementation (which doesn't support new keyring format).

mischelebuha:
I have the same problem, aptly should support gpg 2.1 as soon as possible. Because, finally, this is since debian stretch the common version of gpg.

pszynk:
For me gpg 2.1 works, but I've had a problem with `-passphrase` option (it asked for password anyway).   This answer helped me - [superuser.com](https://unix.stackexchange.com/a/415064/277109).

brucepc:
I have the same problem with `gpg 2.1.18` and `libgcrypt 1.7.6-beta` the option `-passphrase` don't work as planned, with or without `-batch[=true]`.

apachelogger:
Some quick notes for future reference:

I think the `gpg2` provider probably just needs some testing to figure out the necessary parameters (most definitely `--passphrase-file` must be combined with `--batch`. I am not entirely sure how that will play with potential pinentry configuration on the agent side though. Certainly needs some investigation.

[...]

PR #734, opened 2018-04-25
introduce a gpg and gpgv version compatibility check and fall back to v1

apachelogger:
Newer versions of debian and ubuntu come with gpg pointing to gpg2.
We can currently only handle gpg1 CLIs though. Luckily the old gpg is still
available in the package gnupg1 (providing bin/gpg1).

As a bit of a stop-gap, until #657 can be resolved properly, we'll detect
the version of bin/gpg. If it is unsuitable we'll fall back and try
bin/gpg1. If neither is found to be suitable the signer/verifier will
not work.

Same applies to gpgv/gpgv1.

I kind of think we should add some documentation about this on the website as well.


apachelogger:
Should be good for review & landing now.

pgp/gnupg.go
```diff
@@ -121,19 +166,43 @@ func (g *GpgSigner) ClearSign(source string, destination string) error {
...
+
 // InitKeyring verifies that gpg is installed and some keys are trusted
 func (g *GpgVerifier) InitKeyring() error {
-	err := exec.Command("gpgv", "--version").Run()
+	cmd, err := findGPG1()
 	if err != nil {
-		return fmt.Errorf("unable to execute gpgv: %s (is gpg installed?)", err)
+		return err
+	}
+	g.gpg = cmd
```
sliverc: what is the reason that g.gpg and below g.gpgv is reinitialized again here? Is this not already done in NewGpgVerifier?

Issue #741, opened 2018-05-16
Debian packaging for debian stretch


jgeysens:
## Detailed Description
The aptly debian package is out-of-date and that is problematic for debian stretch.

## Possible Implementation
Make sure the debian package installs correctly on debian stretch.  It should depend on the gnupg1 and gpgv1 packages for Debian stretch (breaking change in aptly version 1.3.0).

Even after installing both packages, keys that worked before will not longer work after the upgrade to 1.3.0 (since keys would typically be generated with the default gpg version in stretch - it worked in 1.2.0, so this is a regression)

## Context
`ERROR: unable to initialize GPG signer: looks like there are no keys in gpg, please create one (official manual: http://www.gnupg.org/gph/en/manual.html)`

## Your Environment
Debian Stretch + Aptly 1.3.0 (upgrade from 1.2.0)

jgeysens:
Suggestion : having a package for stretch would solve the dependency issue.

> Second issue is interesting - so signing worked for you in 1.2.0 with default gpg (GnuPG 2.x), but as 1.3.0 falls back to gpg1, gpg1 can't read 2.x keyring?

Correct.  I currently reverted aptly to version 1.2.0 - I don't want to mess up our current configuration, and re-create the keys.

Linking issue #740

apachelogger:
@smira WRT to the second issue, that is an expected side effect of restricting the version. We were semi-compatible with gpg2 iff the gpg implementation was used and iff the interaction extends no further than auth-less signing (e.g. passphrase-less key). Or perhaps a truer statement is: gpg2 CLI is semi-compatible with gpg1 CLI, so we were too :smile:. gpg1 however has no notion or understanding of gpg2 (it's a one-way street, gpg2 imports gpg1 keyrings once and then they go their sepearate ways, gpg1 doesn't do anything of the sort with gpg2). So, essentially what is happening is if you only ever used gpg2 and it happend to _just_ work as per the restriction I mentioned above, then upgrading to 1.3.0 and thus going back to gpg1 means you have no keys (because the keys are in gpg2's keyring and gpg1's will be empty). There isn't really much to be done about this from our side, except for actually building gpg2 support.

Now that I wrote that, perhaps we should put a note about this in the release notes, and tell people that they need to export their keys from gpg2 and import into gpg1 if they happen to have used only gpg2 before?

apachelogger:
[...]

PR #743, opened 2018-05-24
Skip GPG version check `APTLY_SKIP_GPG_VERSION_CHECK=1` is set in the env

smira:
This allows to force using GnuPG 2.x even if aptly is not 100% ready
to use it.

See #741, #657

## Checklist

- [ ] documentation updated

### DIFF
--- a/pgp/gnupg.go
+++ b/pgp/gnupg.go
@@ -18,6 +18,9 @@ var (
 	_ Verifier = &GpgVerifier{}
 )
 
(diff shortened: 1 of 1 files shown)


smira:
This is only first step, another is updating docs with options on using GPG 1.x/2.x

smira:
@apachelogger it might be easier to disable this by setting it to `no` vs. unsetting variable (say in complicated scenarios, when using make). I might have stayed on the safe side here too much :)

PR #744, opened 2018-05-24
Move release build to Travis CI

smira:
This updates previous work in #739 to build
Debian packages and zip files for other OS.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,11 +36,6 @@ endif
 install:
 	go install -v -ldflags "-X main.Version=$(VERSION)"
 
-build:
-	rm -rf build
-	mkdir -p build
-	go build -v -ldflags "-X main.Version=$(VERSION)" -o "build/aptly-$(VERSION)"
-
 system/env: system/requirements.txt
 ifeq ($(RUN_LONG_TESTS), yes)
 	rm -rf system/env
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -66,19 +61,25 @@ mem.png: mem.dat mem.gp
...
-goxc:
+goxc: dev
 	rm -rf root/
 	mkdir -p root/usr/share/man/man1/ root/etc/bash_completion.d/ root/usr/share/zsh/vendor-completions/
 	cp man/aptly.1 root/usr/share/man/man1
 	cp completion.d/aptly root/etc/bash_completion.d/
 	cp completion.d/_aptly root/usr/share/zsh/vendor-completions/
 	gzip root/usr/share/man/man1/aptly.1
-	goxc -pv=$(VERSION) -max-processors=4 $(GOXC_OPTS)
+	goxc -pv=$(VERSION) -max-processors=2 $(GOXC_OPTS)
```
apachelogger: Is there a reason this is hardcoded to 2 rather than get the cpu count from `nproc`?

smira: This is more for local dev, I think `go` does parallel builds on its own, so combined parallelization of goxc + go almost freezes my laptop during the builds. Travis I think doesn't have many procs as well ;)

smira:
Trying to establish flow of artifacts from Travis to S3 bucket (this seems to be working fine).

smira:
@sliverc yep, that's same spurious failures on S3/Swift tests I has never been able to track down

PR #753, opened 2018-06-21
dep: use official uuid package

aviau:
I was taking a look at Debian's official Aptly package and I have noticed that it bundles a ton of dependencies that are already in Debian.

While working on this, I have noticed that aptly uses a mirror of the uuid package, but it turns out that this package was moved to Github so there should be no need for it.

This would make it easier for us in Debian since we wouldn't have to patch every import line...


aviau:
Thanks!

PR #755, opened 2018-07-04
Unit tests for PGP signing/verification

smira:
These unit-tests cover operations via both PGP providers:
built-in `openpgp` and external `gpg`.

Next step is to run these tests for gpg1 & gpg2
as separate entities.

## Checklist

- [ ] documentation updated


smira:
Pushed a small bug fix for clearsign tests with passphrase - they were testing no-passphrase case instead

PR #759, opened 2018-07-13
[RFC] make binary packages optional

### LINKED ISSUE #758: not working with mirrors containing just sources (dsc)
If a mirror just contains source packages, aptly is not able to update it.

aptly always expects that there are binary packages on the mirror.

manut:
this would fix #758 

Signed-off-by: Manuel Traut <manut@linutronix.de>

### DIFF
--- a/cmd/mirror_create.go
+++ b/cmd/mirror_create.go
@@ -18,6 +18,7 @@ func aptlyMirrorCreate(cmd *commander.Command, args []string) error {
 	}
 
 	downloadSources := LookupOption(context.Config().DownloadSourcePackages, context.Flags(), "with-sources")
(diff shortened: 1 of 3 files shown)


manut:
if this solution is basicly acceptable i would fixup the unit-tests, man-pages, etc. just let me know!

sliverc:
In general the approach looks OK to me. @smira might have some more insights though.

smira:
Looks good, except that I would probably invert the logic to `-skip-debs`.

But after cursory look - would it be sufficient to simply supply empty list of architectures?

lbolla:
@neolynx @randombenj I would suggest we close this PR and just add a note in the documentation to say that if we want to work with source-only repos, users just specify no architectures.

Agree?

Issue #761, opened 2018-07-19
db cleanup uses a large amount of memory


rhowe-gds:
On a machine with seven repositories, each containing approximately 3,000 packages except for one with 19,000, aptly db cleanup uses a large amount of memory (>8GB). There are also approximately 10,000 snapshots and ~20 mirrors.

smira:
`aptly db cleanup` is expected to use some amount of memory as it has to load list of all the package IDs into memory, but gigabytes is certainly much more than I would ever expect. Any chance you could have shared your DB in private @rhowe-gds? If not, I can still look into the code to see if there's any room for improvement.

rhowe-gds:
It's about 1.4GB and has been around for about 4 years

smira:
I've got a small fix in #762, I would expect it might take down memory usage 50% or less.

karras:
We're experiencing the same issue. I recently updated aptly from version 0.9.5 to 1.3. As in the past the DB wasn't compressed we have around 2k of DB files summing up to 8GB data. Now `aptly db cleanup` runs into OOM and is killed on a machine with 8GB RAM.

karras:
@sliverc built a version with #762 for me but sadly in our case the cleanup still ran out of memory.

smira:
@rhowe-gds @karras any chance you can try `db cleanup` with latest nightly build, `aptly_1.3.0+30+g72ff71f` or later one?

PR #762, opened 2018-07-19
Lower memory usage for `aptly db cleanup`

smira:
## Description of the Change

This is not a complete fix, but the easiest first step.

During `db cleanup`, aptly is loading every repo/mirror/... into memory,
and even though each object is processed only once, collection holds
a reference to all the loaded objects, so they won't be GC'd until
process exits.

CollectionFactory.Flush() releases pointers to collection objects,
making objects egligble for GC.

This is not a complete fix, as during iteration we could have tried
to release a link to every object being GCed and that would have
helped much more.

See #761

## Checklist

- [ ] documentation updated

### DIFF
--- a/cmd/db_cleanup.go
+++ b/cmd/db_cleanup.go
@@ -59,6 +59,8 @@ func aptlyDbCleanup(cmd *commander.Command, args []string) error {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


smira:
I will merge this PR and I will start working on a more complete solution at the same time.

PR #765, opened 2018-08-02
Implement lazy iteration (ForEach) over collections

smira:
See #761

## Description of the Change

aptly had a concept of loading small amount of info per each object
into memory once collection is accessed for the first time.

This might have simplified some operations, but it doesn't scale well
with huge aptly databases.

This is just intermediate step towards better memory management -
list of objects is not loaded unless some method is called.
`ForEach` method (mainly used in cleanup) is reimplemented to
iterate over database without ever loading all the objects into memory.

Memory was even worse with previous approach, as for each item usually
`LoadComplete()` is called, which pulls even more data into memory
and item stays in memory till the end of the iteration as it is referenced
from `collection.list`.

For the subsequent PR: reimplement `ByUUID()` and probably other methods
to avoid loading all the items into memory, at least for all the collecitons
except for published repos. When published repository is being loaded, it
might pull source local repo which in turn would trigger loading for all the
local repos which is not acceptable.

## Checklist

- [ ] documentation updated

### DIFF
--- a/deb/local.go
+++ b/deb/local.go
@@ -99,28 +99,34 @@ type LocalRepoCollection struct {
 
 // NewLocalRepoCollection loads LocalRepos from DB and makes up collection
 func NewLocalRepoCollection(db database.Storage) *LocalRepoCollection {
-	result := &LocalRepoCollection{
+	return &LocalRepoCollection{
 		RWMutex: &sync.RWMutex{},
 		db:      db,
 	}
+}
 
-	blobs := db.FetchByPrefix([]byte("L"))
-	result.list = make([]*LocalRepo, 0, len(blobs))
+func (collection *LocalRepoCollection) loadList() {
+	if collection.list != nil {
+		return
+	}
+
+	blobs := collection.db.FetchByPrefix([]byte("L"))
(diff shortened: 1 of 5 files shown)


smira:
@sliverc good point, I will try to add a micro-benchmark for it. At least we should see some number, not as good as real test, but getting closer

smira:
@sliverc I did a very simple benchmark (going to push as part of this PR) which does `ForEach()` over collection of 1024 snapshots (could be any other object, code is pretty similar).

Results, before the change (on `master`):

```
BenchmarkSnapshotCollectionForEach-8   	     500	   2790150 ns/op	 1352170 B/op	   30743 allocs/op
```

With this PR:

```
BenchmarkSnapshotCollectionForEach-8   	     500	   2769661 ns/op	 1066171 B/op	   29712 allocs/op
```

I would call this a good result, as there's almost no change in performance (which is expected, as `ForEach` has to go through every snapshot in the end).

For PR #762 I would add more benchmarks which assess performance of methods like `ByName` and `ByUUID` which should have a different performance profile.

sliverc:
@smira :+1:  I agree this result looks good. I don't think this small performance hit will be felt during normal operation.

PR #766, opened 2018-08-06
 Reimplement DB collections for mirrors, repos and snapshots

smira:
See #765, #761

Collections were relying on keeping in-memory list of all the objects
for any kind of operation which doesn't scale well the number of
objects in the database.

With this rewrite, objects are loaded only on demand which might
be pessimization in some edge cases but should improve performance
and memory footprint significantly.

This doesn't touch `PublishedRepoCollection` as it relies on list of 
all the objects in many places to implement unique checks, proper
cleanup.

## Checklist

- [ ] documentation updated

### DIFF
--- a/deb/local.go
+++ b/deb/local.go
@@ -2,6 +2,7 @@ package deb
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"log"
 	"sync"
@@ -93,8 +94,8 @@ func (repo *LocalRepo) RefKey() []byte {
 // LocalRepoCollection does listing, updating/adding/deleting of LocalRepos
 type LocalRepoCollection struct {
 	*sync.RWMutex
-	db   database.Storage
-	list []*LocalRepo
+	db    database.Storage
+	cache map[string]*LocalRepo
 }
 
 // NewLocalRepoCollection loads LocalRepos from DB and makes up collection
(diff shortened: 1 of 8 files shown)


deb/local.go
```diff
@@ -175,26 +186,39 @@ func (collection *LocalRepoCollection) LoadComplete(repo *LocalRepo) error {
...
+	value, err := collection.db.Get(key)
+	if err == database.ErrNotFound {
+		return nil, fmt.Errorf("local repo with uuid %s not found", uuid)
+	}
+
+	if err != nil {
+		return nil, err
+	}
+
+	r := &LocalRepo{}
```
sliverc: this doesn't seem to be covered by a test (https://codecov.io/gh/aptly-dev/aptly/pull/766/diff#D2-210)

I think as this is a success path it would be good to have a unit test for it. There are some similar code bits uncovered in the other collections too.

smira: yep, thanks, I will check it!

smira: I've improved test coverage and added some benchmarks for `ByUUID()` and `ByName()`. They exercise "worst case": collection is created and one lookup is performed. On `master` branch this always leads to loading all the objects. 

For `ByUUID()` in the new approach loading is fast, as object is looked up directly (loading only single element). `ByName()` still requires scanning whole collection, in the worst case it is as bad as the old approach, on average it's 50% better.

`ForEach()` doesn't cache objects in memory, so this should help #761 as objects would be GCed as soon as they're scanned. `ByUUID()` is used a lot to lookup source of published repositories.

On `master`:

```
BenchmarkSnapshotCollectionByUUID-8    	     500	   2953932 ns/op	 1352168 B/op	   30743 allocs/op
BenchmarkSnapshotCollectionByName-8    	     500	   2922043 ns/op	 1352504 B/op	   30747 allocs/op
```

This branch:

```
BenchmarkSnapshotCollectionByUUID-8    	  300000	      4492 ns/op	    1792 B/op	      39 allocs/op
BenchmarkSnapshotCollectionByName-8    	    1000	   1433058 ns/op	  533994 B/op	   14870 allocs/op
```

smira: I also had an idea to implement very simple "index" to lookup things by name (which is really frequent lookup): just a key in DB which is like `name` -> `UUID`. `ByName()` could use this index optimistically - if it's missing or doesn't point to right entry, `ByName()` falls back to full scan.

sliverc: This could be a way. I think we should see first how these changes are in terms of performance in a actual use case before we make the code even more complex.

PR #801, opened 2019-01-03
Fix #800: -skip-existing-packages for missing checksums

### LINKED ISSUE #800: -skip-existing-packages not working when remote does not define all checksums
## Detailed Description
Some debian package servers do not send the `MD5Sum` field in their package list.
This is a [known](#228) issue and was [fixed](#539 ) in release 1.1.0. :+1:

However when you update the mirror of such a server the `-skip-existing-packages` is currently not working, as it relies on sha1, md5 and sha256 being present to calculate the filehash.

## Possible Implementation
I have a PR ready which will look the packages up in the package pool when `-skip-existing-packages` is present, but not all checksums are available.

steven-aerts:
Skipping of existing packages will now also work when not all checksums
are provided by the mirrored repository.
When this is the case, packages will be searched for in the package
pool.  And the search results will be validated against the requested
package.

Fixes #800 

I am aware this code does not contain any tests.  I am scheduling it for review first to see if my proposed solution is acceptable.  If it is, I can add some tests.

## Checklist

- [ ] documentation updated

### DIFF
--- a/deb/reflist.go
+++ b/deb/reflist.go
@@ -69,7 +69,7 @@ func (l *PackageRefList) Encode() []byte {
 	return buf.Bytes()
 }
 
-// Decode decodes msgpack representation into PackageRefLit
+// Decode decodes msgpack representation into PackageRefList
 func (l *PackageRefList) Decode(input []byte) error {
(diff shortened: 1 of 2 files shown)


lbolla:
This looks like a workaround for broken repositories. Not worth merging imo.

PR #807, opened 2019-01-19
Don't remove API file socket if it exists and usable

### LINKED ISSUE #806: Unable to set permissions on the unix socket
## Detailed Description

If you use the unix socket to serve the aptly api (using `aptly serve --listen="unix:///var/run/aptly.sock"`) , you are not able to set the permissions on the socket file.

The file is deleted each time it starts, and recreated (see [here](https://github.com/aptly-dev/aptly/blob/9a704de43b0937b55b90dc0c6bbd1591a3b20203/cmd/api_serve.go#L61)), so the permissions are resetted at start.

Using a post command to set the permission on the socket file does not work (with Systemd `ExecStartPost=/bin/chmod g+w /var/run/aptly.sock`). This is probably because the socket file is deleted asynchronously (after the post command run)

## Possible Implementation

I think we just need to not delete the socket file. So an error should be returned if the socket cannot be created, and use it if it already exsits. It should work because [the socket is created with the syscall.SO_REUSEADDR option](https://golang.org/src/net/sockopt_linux.go).

By this way, we are able to set the permissions after and it will not reset them when the service start.

Or

Set the permission with the fragment part of the URL `unix:///var/run/aptly.sock#660` :)

After the socket file is created:
```
err := os.Chmod(file, listenURL.Fragment);
if err != nil {
   return fmt.Errorf("unable to serve: %s", err)
}
```

## Your Environment
Debian 9

Thanks !

smira:
Fixes #806

Also updates linter config & fixes linting errors.

### DIFF
--- a/cmd/api_serve.go
+++ b/cmd/api_serve.go
@@ -58,7 +58,25 @@ func aptlyAPIServe(cmd *commander.Command, args []string) error {
 	listenURL, err := url.Parse(listen)
 	if err == nil && listenURL.Scheme == "unix" {
 		file := listenURL.Path
-		os.Remove(file)
+
+		var stat os.FileInfo
+		stat, err = os.Stat(file)
+		shouldRemove := true
+
+		if err == nil && stat.Mode()&os.ModeSocket == os.ModeSocket {
+			shouldRemove = false
+		}
+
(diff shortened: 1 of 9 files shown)


smira:
Merging this to fix build failures

PR #831, opened 2019-04-15
Allow a custom Suite when publishing

rul:
## Description of the Change

A `-suite` parameter is added for publishing. If it's not provided, then `distribution` is used (current default).

Why this change is important?

Starting from Ubuntu bionic, `apt` checks that the `Label`, `Origin` and `Suite` of a repo remains the same. If they change, it requires manual confirmation from the user. Hence, administrators migrating to `aptly` need to have the flexibility to set a custom Suite field.

This change is also important because distributions use the Suite for tagging the release in different ways. For example, Debian uses the suites for tagging `stable`, `oldstable`, `oldoldstable`, etc.

## Checklist

- [ ] documentation updated

# About tests
At the moment, not all the tests are passing. I think this might be related to the [recent archival](https://lists.debian.org/debian-devel-announce/2019/03/msg00006.html) of wheezy. I added tests related to this change, but didn't fix the ones that were already broken.


smira:
Closing this PR, #874 is rebased version of it.

PR #854, opened 2019-07-10
Try Travis on xenial workers

smira:


smira:
Coverage going down, as tests don't work with GnuPG 2.1.x from xenial, but they work with either 2.0.x from trusty or 2.2.x from bionic+, but bionic has its own pile of issues.

Issue #791, opened 2018-10-26
Error decoding mirror: msgpack decode error


pfelecan:
Since the last update on a Debian testing, 1.3.0+ds1-1 from 1.3.0-6, nothing works anymore.
For example:
```
aptly mirror list
Error decoding mirror: msgpack decode error [pos 200]: invalid length of bytes for decoding time - expecting 4 or 8 or 12, got 15
Error decoding mirror: msgpack decode error [pos 190]: invalid length of bytes for decoding time - expecting 4 or 8 or 12, got 15
No mirrors found, create one with `aptly mirror create ...`.
```
There are mirrors, two of them.

I tried an `aptly db recover`, to no avail.

pfelecan:
I reverted the installation, i.e. removed the new package and reinstalled the old one.
The aforementioned error does not appear anymore.
Consequently, it's a packaging/build error.
Time to open a bug report on the Debian tracker.

smira:
Yep, aptly was built without provided `vendor/`, and `Gopkg.lock` wasn't used as well. 

You can use builds from https://www.aptly.info/download/

pfelecan:
This is quite annoying.
What the Debian packager/maintainer recommends is to "recreate" the database.
Is that possible? How can we recreate the database?

smira:
There's only one way -  drop the database (`~/.aptly/db`) and aptly will create new one, but this will lead to all the aptly data being *erased*.

smira:
What is even worse - DB would be incompatible with aptly builds from repo.aptly.info.

smira:
The best thing I can do is to provide fixed version of aptly which will enforce older encoding with new version of the codec but Debian package should be update

smira:
Workaround is described in ugorji/go#269

taladar:
Wouldn't it make sense to have migration code from old versions of the DB time encoding to new versions? aptly will have to move to the new version of the dependency eventually, especially since we are talking serialization/deserialization libraries and those (in general) tend to be notorious for security issues.

pfelecan:
FYI, I opened an issue on the Debian tracker 911924

smira:
@taladar I'm talking about moving to new version of the library but configuring it in a way to read and write data in the same format as the vendored version today in aptly

smira:
Looks like new version of ugorji/codec is even more broken then I might have assumed. It skips fields tagged as `json:"-"` for msgpack encoding.

neolynx:
[...]

PR #830, opened 2019-04-13
Update ugorji/go/codec to latest version

zhsj:
## Requirements

All new code should be covered with tests, documentation should be updated. CI should pass.

## Description of the Change

The vendored ugorji/go/codec is rather old...

## Checklist

- [ ] unit-test added (if change is algorithm)
- [ ] functional test added/updated (if change is functional)
- [ ] man page updated (if applicable)
- [ ] bash completion updated (if applicable)
- [ ] documentation updated
- [x] author name in `AUTHORS`

PR #855, opened 2019-07-12
Update ugorji/go/codec to latest version

smira:
This is re-post of PR #830 rebased to the latest master


smira:
This should fix #791, but that requires some verification

smira:
This seems to have backwards compatibility and also provides a way to build with latest variant of msgpack codec.

PR #861, opened 2019-08-01
Refactor database code to support standalone batches, transactions.

smira:
This is spin-off of changes from #459.

Transactions are not being used yet, but batches are updated to work
with the new API.

`database/` package was refactored to split abstract interfaces and
implementation via goleveldb. This should make it easier to implement
new database types.

## Checklist

- [ ] documentation updated


smira:
> Could you move the changing of mirror into another PR? Otherwise this makes reviewing harder as there are code and test changes which are not relevant to each other.

Yep, makes sense. I will split. I was trying to fix the build in Travis CI, but looks like the failures are still there. Going to add retries to aptly on network errors.

> In terms of database interface - do we want to support another database anytime soon?

That was the initial plan.

> In my opinion in golang interface should only be defined if there are actually used as that's the power of golang that those can be define at any time fairly easily. So if we don't implement another database any time soon I would not define interfaces.

Interface was there since the inception of aptly, I just moved things around so that's it more obvious what is interface, and what is the implementation.

I had plans to add other types of databases, in fact it might make sense to support something distributed (Redis?), but this is not my goal for this PR of course.

Interface plays another important role here: it abstracts away goleveldb, so we don't end up using features in a random way, it's clear what's being used and what not.

sliverc:
@smira My quick look over the changes was a bit too quick... :wink: let's leave the interfaces as they were before then.

smira:
Split not related commits into #863, #864

system/t04_mirror/UpdateMirror11Test_gold
```diff
@@ -13,9 +13,6 @@ Downloading ftp://ftp.ru.debian.org/debian/pool/main/s/sensible-utils/sensible-u
 Downloading ftp://ftp.ru.debian.org/debian/pool/main/s/sysvinit/sysvinit-utils_2.88dsf-59.9_i386.deb...
 Mirror `stretch-main` has been successfully updated.
 Packages filtered: 50621 -> 3.
-gpgv: Good signature from "Debian Archive Automatic Signing Key (7.0/wheezy) <ftpmaster@debian.org>"
```
smira: these and the two below should be removed from the PR, will get back once tests are stable enough

smira:
@sliverc updated, should be ready for the review now

PR #863, opened 2019-08-05
Fix flakey tests related to identity name ordering.

smira:
Identities were not sorted, so randomly it might print them in different order breaking tests.

Also added filters to download less data.


smira:
Rebased without changes.

PR #864, opened 2019-08-05
Print redirects being followed, drop mirror.yandex.ru.

smira:
Use CDN-backed Debian mirror to make tests run faster hopefully for
everyone. Redirects might be important to know what exactly is going on
when items are being downloaded.


smira:
Rebased without changes

PR #865, opened 2019-08-05
Rework HTTP downloader retry logic

smira:
Apply retries as global, config-level option `downloadRetries` so that
it can be applied to any aptly command which downloads objects.

Unwrap `errors.Wrap` which is used in downloader.

Unwrap `*url.Error` which should be the actual error returned from the
HTTP client, catch more cases, be more specific around failures.


smira:
Travis issue discussion: https://travis-ci.community/t/travis-network-timeout-issue/4584

smira:
Tests are still failing which I believe is related to the Travis issue, but while at it, I refactored download retries, and now they can be enabled via config.

man/Makefile
```diff
@@ -1,10 +1,12 @@
-all: prepare generate
+RUBYBINPATH=$(shell ruby -r rubygems -e 'puts Gem.user_dir')/bin
```
smira: with that, `make man` runs under regular user account without access to system Ruby lib directories

smira:
Merging this to unblock other PRs, and this also makes retries work much more reliably.

PR #880, opened 2019-09-26
Switch to Go modules instead of dep

smira:
As minimum support Go version for aptly is 1.11, we can start using Go modules.

This makes repo much more slim (no `vendor/` directory), and provides better support going forward to use mainstream dependency management.


smira:
Fails due to #881

PR #919, opened 2020-04-29
Waiting for S3 to become consistent when moving files

### LINKED ISSUE #374: Updating on S3 needs mutual exclusion on API
When a repository is being updated to S3, if there is another thread doing the same action, last action can fails.

The following log show previous case:

```
[GIN] 2016/04/01 - 07:39:26 | 200 |  724.63868ms | 10.50.30.251:58859 |   POST    /api/files/folder1
[GIN] 2016/04/01 - 07:39:26 | 200 | 555.043702ms | 10.50.30.251:58860 |   POST    /api/repos/repository/file/folder1
Signing file 'Release' with gpg, please enter your passphrase when prompted:
Clearsigning file 'Release' with gpg, please enter your passphrase when prompted:
[GIN] 2016/04/01 - 07:39:43 | 200 | 17.203494609s | 10.50.30.251:58861 |   PUT     /api/publish/s3:repo:/precise
[GIN] 2016/04/01 - 07:39:52 | 200 | 1.324651495s | 10.50.30.251:58862 |   POST    /api/files/folder2
[GIN] 2016/04/01 - 07:39:53 | 200 | 950.391809ms | 10.50.30.251:58863 |   POST    /api/repos/repository/file/folder2
Signing file 'Release' with gpg, please enter your passphrase when prompted:
Clearsigning file 'Release' with gpg, please enter your passphrase when prompted:
[GIN] 2016/04/01 - 07:40:10 | 500 | 17.75603111s | 10.50.30.251:58864 |   PUT     /api/publish/s3:repo:/precise
Error #01: unable to update: unable to rename: error copying dists/precise/main/binary-amd64/Release.tmp -> dists/precise/main/binary-amd64/Release in S3: us-east-1:repo/: NoSuchKey: The specified key does not exist.
    status code: 404, request id: 164EAA42579D5CE2 
     Meta: Operation aborted
```

I guess this error should happen even updating on disk.

Adding a mutual exclusion on PUT could fix this problem.

Aptly version: 0.9.7~dev

marcinswigon:
Fixes #374


lbolla:
We try not to sleep.

marcinswigon:
@lbolla Will you provide another solution for S3 synchronization issues?

lbolla:
@marcinswigon The thing is that sleeping is not guaranteed to work: which is the main reason why I closed this PR.

The original issue #374 was 4 years old and got closed during the "grooming" of old tickets. I can see your recent comment in that issue, but I am not sure it's related to the original report. The original report had to do with two thread publishing on the same bucket and the second failing, whereas yours seems to have to do with "eventual consistency" of S3 buckets.

So maybe I am misreading the issue. Can you elaborate more?

marcinswigon:
I have looked into AWS docs and apparently in December 2020 they have introduced strong Read-After-Write consistency model: https://aws.amazon.com/blogs/aws/amazon-s3-update-strong-read-after-write-consistency/. That bug appeared only in the old eventual consistency model (probably that's why we did not see it appearing for a while).

lbolla:
@marcinswigon Great! So, the right solution is probably on the S3 side, to configure the bucket with "read-after-write". Aptly can't do much about it. Agree to close?

marcinswigon:
Sure. I wasn't aware that S3 changed consistency model. This issue should not happen anymore :)

lbolla:
Thanks!

PR #924, opened 2020-07-22
Store MD5 in a separate metadata field and use it when ETag is not an MD5 hash

### LINKED ISSUE #923: S3 backend relies on ETag being MD5
```go
md5s = append(md5s, strings.Replace(*key.ETag, "\"", "", -1))
```

## Detailed Description
The S3 backend relies on ETag S3 returns being equal to the MD5 of the object, but it’s not necessarily true. See https://docs.aws.amazon.com/AmazonS3/latest/API/RESTCommonResponseHeaders.html:

> The entity tag is a hash of the object. The ETag reflects changes only to the contents of an object, not its metadata. The ETag may or may not be an MD5 digest of the object data. Whether or not it depends on how the object was created and how it is encrypted as described below:
> 
> Objects created by the PUT Object, POST Object, or Copy operation, or through the AWS Management Console, and are encrypted by SSE-S3 or plaintext, have ETags that are an MD5 digest of their object data.
> 
> Objects created by the PUT Object, POST Object, or Copy operation, or through the AWS Management Console, and are encrypted by SSE-C or SSE-KMS, have ETags that are not an MD5 digest of their object data.
> 
> If an object is created by either the Multipart Upload or Part Copy operation, the ETag is not an MD5 digest, regardless of the method of encryption.

(Originally posted by @harshavardhana in minio/minio#9999)

Instead, Aptly should store relevant checksums in the associated metadata.

andrewshadura:
## Description of the Change

The S3 backend relies on ETag S3 returns being equal to the MD5 of the object, but it’s not necessarily true. For that purpose we store the MD5 object in a separate metadata field as well to make sure it isn’t lost. When the value returned clearly doesn’t look like a valid MD5 hash (length isn’t exactly 32 characters), attempt to retrieve the MD5 hash possibly stored in the metadata.

We cannot always do this since user-defined metadata isn’t returned by the ListObjects call, so verifying it for each object is expensive as it requires one HEAD request per each object.

This pull request fixes #923

## Requirements

## Checklist

- [ ] documentation updated

### DIFF
--- a/s3/public.go
+++ b/s3/public.go
@@ -150,6 +150,20 @@ func (storage *PublishedStorage) PutFile(path string, sourceFilename string) err
 	return err
 }
 
+// getMD5 retrieves MD5 stored in the metadata, if any
+func (storage *PublishedStorage) getMD5(path string) (string, error) {
+	params := &s3.HeadObjectInput{
+		Bucket: aws.String(storage.bucket),
+		Key:    aws.String(filepath.Join(storage.prefix, path)),
+	}
+	output, err := storage.s3.HeadObject(params)
+	if err != nil {
+		return "", err
+	}
+
+	return aws.StringValue(output.Metadata["Md5"]), nil
+}
+
 // putFile uploads file-like object to
 func (storage *PublishedStorage) putFile(path string, source io.ReadSeeker, sourceMD5 string) error {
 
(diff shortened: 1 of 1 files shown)


andrewshadura:
Ping?

s3/public.go
```diff
@@ -298,6 +317,17 @@ func (storage *PublishedStorage) LinkFromPool(publishedDirectory, fileName strin
 	sourceMD5 := sourceChecksums.MD5
 
 	if exists {
+        if len(destinationMD5) != 32 {
```
lbolla, line 320: This seems wrongly indented. Please check with "go fmt".

s3/public.go
```diff
@@ -142,16 +142,30 @@ func (storage *PublishedStorage) PutFile(path string, sourceFilename string) err
 	}
 	defer source.Close()
 
-	err = storage.putFile(path, source)
+	err = storage.putFile(path, source, "")
```
lbolla, line 145: Probably I am misunderstanding the issue, but shouldn't we calculate MD5 of the content here and use it when putting the file?

andrewshadura: As far as I remember, hashes matter only for files from the pool, since the lists are recreated anyway; files in the pool are only being uploaded from `LinkFromPool`, so I left this part without a change.

Of course, it would be a bit more future-proof in case the rest of the code changes and pool files start being uploaded here too, but right now it would be calculating MD5 only for that value to never being used again.

s3/public.go
andrewshadura: Ack, updating,

lbolla:
Can you also rebase against the latest master please, so tests should now pass.

lbolla (approved):

PR #936, opened 2021-01-08
deb: fix importing dbgsym packages with versioned Source field

peat-psuwit:
## Requirements

## Description of the Change

dpkg-gencontrol can be called with -v flag which set binary package's
version separated from source version. When this happen, the Source
field will contain version number in addition to source package name.
This tripped Aptly's dbgsym restriction, which check for exact source
package name, which in turn prevents the dbgsym & the whole .changes
file from being imported.

From the git history, it seems like this condition is a leftover from
when Aptly filter dbgsym packages using "*-dbgsym". So, I decided to
remove it. A test case has been added to prevent regression.

## Checklist

- [ ] documentation updated


peat-psuwit:
There seem to be some PGP failures due to key expiration/revocation.

peat-psuwit:
The commit is rebased. PTAL.

mariogrip:
ping @lbolla

mariogrip (approved):
LGTM :+1:

deb/changes.go
```diff
@@ -195,11 +195,6 @@ func (c *Changes) PackageQuery() PackageQuery {
 			}
 		}
 
-		ddebQuery = &AndQuery{
```
lbolla, line 198: Doesn't removing filtering by `c.Source` potentially select the wrong packages?

peat-psuwit: We already have `<pkg name>-dbgsym` rule for each specific package, and it's not like multiple source packages should produce the same binary package. `binaryQuery` doesn't have `c.Source` rule, and it's constructed essentially the same way as `ddebQuery`.

lbolla:
@peat-psuwit if you rebase this against master I can then merge.

peat-psuwit:
Please take another look. The test failure seems to be external.

peat-psuwit:
@lbolla please take another look.

lbolla:
@neolynx What do you think of this patch? I am not too familiar with this part: I'd appreciate your input before merging.

lbolla (commented):
@peat-psuwit Would you mind rebasing this once more? master now has fully functional unittests run on GH Actions, whose successful run is required for each PR. Thanks!

neolynx:
LGTM

peat-psuwit:
Rebased and adapt to a change in test utils. Please have another look.

randombenj (approved):
LGTM

lbolla (approved):

PR #937, opened 2021-02-06
Fix tests

### LINKED ISSUE #933: build failing
The readme indicates the build has been failing for many months.

lbolla:
Fixes #933 

Fix tests failing on Travis.
Most of the tests were failing because of expired PGP keys. I've downloaded the most recent ones from Debian (and elsewhere), updated the fixtures and fixed some more tests.

The patchset is already large as it is. In following PRs I'd like to:
* test against newer versions of Go
* add support for Python3 (and maybe drop Python2?)
* test against Bionic/Focal instead of Xenial


lbolla:
Interesting, [normal Travis tests pass](https://travis-ci.org/github/aptly-dev/aptly/builds/757832458?utm_source=github_status&utm_medium=notification) but [the ones triggered by a PR](https://travis-ci.org/github/aptly-dev/aptly/builds/757838687?utm_source=github_status&utm_medium=notification) do not.
I'll look into why.

szakalboss (approved):

lbolla:
Simply re-running the failing Travis job above made it pass. So, the problem must be some race condition or otherwise aleatory condition. I think it makes sense to merge this patch and tackle this other problem in another PR.

PR #949, opened 2021-03-29
Flat repos

Smarre:
This supersedes #434. For now this is just rebased from master.

Please note that this is complete work in progress, and still shares same issues as parent MR. This is a draft for those reasons. Do not merge this now.

In addition to those minor things I found, there is few big issues that needs to be resolved:

- [ ] Existence of Packages file needs to be checked, to not return false positivies for nonexistent directories.
- [ ] system test with flat repository which does mirror update to make sure there's no regression for flat repositories in the future

## Requirements

## Description of the Change

Continuing work in earlier pull request.

## Checklist

- [ ] documentation updated


system/t04_mirror/CreateMirror6Test_gold
```diff
@@ -1,3 +1,5 @@
 Downloading http://cdn-fastly.deb.debian.org/debian/dists/suslik/InRelease...
 Downloading http://cdn-fastly.deb.debian.org/debian/dists/suslik/Release...
 ERROR: unable to fetch mirror: HTTP code 404 while fetching http://cdn-fastly.deb.debian.org/debian/dists/suslik/Release
+Mirror [mirror6]: http://mirror.yandex.ru/debian/ suslik successfully added.
```
Smarre, line 4: I wonder where these yandex.ru addresses comes from.

system/t04_mirror/create.py
```diff
@@ -58,9 +58,7 @@ class CreateMirror6Test(BaseTest):
     """
     create mirror: missing release
     """
-    expectedCode = 1
```
Smarre, line 61: I don’t get why this line was removed.

cmd/mirror_update.go
```diff
@@ -52,6 +52,11 @@ func aptlyMirrorUpdate(cmd *commander.Command, args []string) error {
 		return fmt.Errorf("unable to update: %s", err)
 	}
 
+	if len(repo.ReleaseFiles) == 0 {
```
Smarre, line 55: Need to check that all possible locations of Release file is covered here.

deb/remote.go
```diff
@@ -274,6 +274,10 @@ func (repo *RemoteRepo) Fetch(d aptly.Downloader, verifier pgp.Verifier) error {
 		// 0. Just download release file to temporary URL
 		release, err = http.DownloadTemp(gocontext.TODO(), d, repo.ReleaseURL("Release").String())
 		if err != nil {
+			if err1, ok := err.(*http.Error); ok && (err1.Code == 404 || err1.Code == 403) {
```
Smarre, line 277: I’m not sure if this 403 is a good practice. It practically says that there *is* a Release file, one just is not allowed to access it.

deb/remote.go
```diff
@@ -302,6 +306,10 @@ func (repo *RemoteRepo) Fetch(d aptly.Downloader, verifier pgp.Verifier) error {
 		// 2. try Release + Release.gpg
 		release, err = http.DownloadTemp(gocontext.TODO(), d, repo.ReleaseURL("Release").String())
 		if err != nil {
+			if err1, ok := err.(*http.Error); ok && (err1.Code == 404 || err1.Code == 403) {
```
Smarre, line 309: Same as above:

I’m not sure if this 403 is a good practice. It practically says that there is a Release file, one just is not allowed to access it.

system/t04_mirror/create.py
Smarre: To answer myself: since this PR is trying to implement support for repositories missing Release, failure should not be expected...

lbolla:
This PR seems like a workaround for broken repos. Closing for now.

PR #955, opened 2021-04-12
Implement support for TPM-backed signing keys (#953)

### LINKED ISSUE #953: Support TPM-backed signing key storage
- Signing package repositories is good.
- Letting the keys used to sign package repositories be copied off of the server authorized for that use is bad.
- Hardware-backed key stores exist to provide a means of storing a private key that prevents it from being copied from software.
- The Trusted Platform Module is a widely-adopted standard that provides for such private storage.
- [Frank Olbricht's `tpmk` library](https://github.com/folbricht/tpmk) exists to allow TPM-backed keys to be operated on from Go.

Aptly should support signing packages with tpmk-backed keys; this way, even a user with all the privileges aptly itself has won't be able to copy that key off and put it to their own uses in the future (but will need to have a persistent compromise of a repo-signing server lasting for the entire duration of time in which they wish to corruptly/incorrectly sign keys).

## Detailed Description
The proposal is to add an alternative to the current practice of passing a signing key (or a path to a signing key) and a password to aptly. For example, we might recognize a `SecretKeyring` of the form `tpm:0x81000000` as an instruction to ask the tpmk to load a private key from the given address.

## Context
This change will make it possible to trust users to use aptly without giving them enough permissions to impersonate that aptly server in the future.

charles-dyfis-net:
Fixes #953

## Requirements

This should not be removed from draft state until the below are checked.

- [x] Documentation is updated
- [ ] Test coverage has been extended to cover the new functionality.
- [x] Switch from the charles-dyfis-net fork of tpmk to upstream (which at the time of that fork's creation did not support clearsigning, but does now).

The test suite shall use Google's TPM emulator for Go (supported by tpmk and used for its own test suite), so it can be run on systems without TPM hardware.

## Description of the Change

Use the Go library github.com/folbricht/tpmk to support TPM-backed keys for OpenPGP signatures.

This allows hardware-backed keys to be used for repository signing. Because the private key data cannot be copied out of the TPM into general-purpose RAM, these keys cannot be stolen by an attacker and used to sign content on a different host; as soon as an attacker's access to the host with the TPM is eliminated, that attacker can no longer generate signatures with any key that TPM stored.

This feature is not available in released versions of upstream GnuPG, so it exclusive to the internal Go implementation.

## Checklist

- [ ] ~~bash completion updated~~ (not applicable)
- [x] documentation updated


charles-dyfis-net:
One point on which I would appreciate feedback: Is the "documentation" a separate thing from the man page? Where, other than the man page, should I be updating documentation?

charles-dyfis-net:
@folbricht has tracked down the build failure -- while tpmk declared a dependency on Go 1.11, it in fact requires Go 1.15.

I expect this PR will need to be extended to hide the feature behind a tag (or otherwise make it conditionally compiled).

charles-dyfis-net:
...a question for the aptly team here: What's the feasibility of having a go 1.15 builder in Travis, so this code can be tested if it's moved behind such a flag?

lbolla:
@charles-dyfis-net Travis is no more. We have GH Actions running Go 1.14, 15, 16, and 17.

Would you like to rebase against master?

randombenj:
I will close this as there seems to be a lack of interest for this change, feel free to leave a comment if you want this reopened

neolynx:
See #1288

PR #959, opened 2021-06-01
Update ParsePrefix to handle prefixes with literal colons

### LINKED ISSUE #958: Cannot publish to location with literal colon in directory name
Over both the CLI and the API, attempts to publish to a location with a `:` character in the directory fail.

## Detailed Description
At present, aptly cannot be used in sites that have directory naming conventions including `:`s. While the use of colons to denote separators inside path specifications does necessarily place restrictions on their use, these restrictions need not, and should not, be fatal.

```
$ aptly publish snapshot -distribution=test -label="qa test" test-releases-1.2.3-qa-final:1.2.3.1 test/releases/1.2.3/qa-final:1.2.3.1
ERROR: unknown published storage format: test/releases/1.2.3/qa-final
$ aptly publish snapshot -distribution=test -label="qa test" filesystem:test-releases-1.2.3-qa-final:1.2.3.1 test-releases/1.2.3/qa-final:1.2.3.1
ERROR: published local storage test/releases/1.2.3/qa-final not configured
$ aptly publish snapshot -distribution=test -label="qa test" filesystem::test-releases-1.2.3-qa-final:1.2.3.1 test-releases/1.2.3/qa-final:1.2.3.1
ERROR: published local storage :test/releases/1.2.3/qa-final not configured
```

...and after configuring a local storage mechanism called `main`:

```
$ aptly publish snapshot -distribution=test -label="qa test" test-releases-1.2.3-qa-final:1.2.3.1 filesystem:main:test/releases/1.2.3/qa-final:1.2.3.1
ERROR: published local storage main:test/releases/1.2.3/qa-final not configured
```

To be clear, removing the colon from the publication location addresses all issues:

```
$ aptly publish snapshot -distribution=test -label="qa test" test-releases-1.2.3-qa-final:1.2.3.1 filesystem:main:test/releases/1.2.3/qa-final--1.2.3.1
Loading packages...
Generating metadata files and linking package files...
Finalizing metadata files...

Snapshot test-releases-1.2.3-qa-final:1.2.3.1 has been successfully published.
```

## Context
We have a convention by which, under our root directory, publications should have locations like `<distro>/versions/<version-series>/<release-type>:<release-spec>`.

Right now, publications of this form cannot be created by aptly.

## Possible Implementation
Given a format in the specification `<type>:<name>:<path>`, everything after the 2nd `:` should be treated as literal text with no special treatment; and it should be possible to leave the `<name>` blank to get the default `~/.aptly/public` publishing location (thus, `filesystem::a:b` to publish to `~/.aptly/public/a:b`).

## Your Environment
Linux (Ubuntu 18.04.4 LTS)

charles-dyfis-net:
Fixes #958

## Requirements

- [X] New unit tests added covering the algorithm change
- [ ] New functional tests added ensuring end-to-end functionality with updated filenames
- [X] No new regressions in the test suite seen (gpg tests broken on my platform, can't get a clean `make test` even with unmodified codebase).

## Description of the Change

Without this, publication locations containing a `:` -- a perfectly legal character in UNIX directory names -- are not possible.

With this change, the following commands work, but did not before:

```
aptly publish snapshot -distribution=test test-releases-1.2.3-qa-final:1.2.3.1 test/releases/1.2.3/qa-final:1.2.3.1
aptly publish snapshot -distribution=test test-releases-1.2.3-qa-final:1.2.3.1 filesystem:test/releases/1.2.3/qa-final:1.2.3.1
aptly publish snapshot -distribution=test test-releases-1.2.3-qa-final:1.2.3.1 filesystem::test/releases/1.2.3/qa-final:1.2.3.1
aptly publish snapshot -distribution=test test-releases-1.2.3-qa-final:1.2.3.1 filesystem:fsname:test/releases/1.2.3/qa-final:1.2.3.1
```

...as do corresponding API calls.

## Checklist

- [ ] documentation updated

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -40,3 +40,4 @@ List of contributors, in chronological order:
 * Raphael Medaer (https://github.com/rmedaer)
 * Raul Benencia (https://github.com/rul)
 * Don Kuntz (https://github.com/dkuntz2)
+* Charles Duffy (https://github.com/charles-dyfis-net)
--- a/context/context.go
+++ b/context/context.go
@@ -343,7 +343,7 @@ func (context *AptlyContext) GetPublishedStorage(name string) aptly.PublishedSto
 
 	publishedStorage, ok := context.publishedStorages[name]
 	if !ok {
-		if name == "" {
+		if name == "" || name == "filesystem" {
 			publishedStorage = files.NewPublishedStorage(filepath.Join(context.config().RootDir, "public"), "hardlink", "")
 		} else if strings.HasPrefix(name, "filesystem:") {
 			params, ok := context.config().FileSystemPublishRoots[name[11:]]
--- a/deb/publish.go
+++ b/deb/publish.go
@@ -9,6 +9,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"strings"
 	"sync"
@@ -70,17 +71,33 @@ type PublishedRepo struct {
 	AcquireByHash bool
 }
 
+var ParsePrefixRegexp *regexp.Regexp
+
+func init() {
+	ParsePrefixRegexp = regexp.MustCompile(`\A(?:(?P<StoreType>[^:/]+)(?:[:](?P<StoreName>[^:/]*))?[:])?(?P<Prefix>.*?)\z`)
+}
+
 // ParsePrefix splits [storage:]prefix into components
+//
+// - `storage` may be of the form `type:name`, so the longest possible form is `type:name:prefix`
+// - because `/`s are only valid in the Prefix field, any `:` after the first `/` must be literal
 func ParsePrefix(param string) (storage, prefix string) {
-	i := strings.LastIndex(param, ":")
-	if i != -1 {
-		storage = param[:i]
-		prefix = param[i+1:]
+	pieces := ParsePrefixRegexp.FindStringSubmatch(param)
+	if param == "" || pieces == nil { // should not ever happen
+		prefix = param
+	} else {
+		storeType := pieces[1]
+		storeName := pieces[2]
+		prefix = pieces[3]
+		if storeType != "" && storeName != "" {
+			storage = storeType + ":" + storeName
+		} else {
+			// only one at most is nonempty; just concatenate
+			storage = storeType + storeName
+		}
 		if prefix == "" {
 			prefix = "."
 		}
-	} else {
-		prefix = param
 	}
 	prefix = strings.TrimPrefix(strings.TrimSuffix(prefix, "/"), "/")
 	return
--- a/deb/publish_test.go
+++ b/deb/publish_test.go
@@ -875,3 +875,24 @@ func (s *PublishedRepoRemoveSuite) TestRemoveRepo5(c *C) {
 	c.Check(filepath.Join(s.publishedStorage2.PublicPath(), "ppa/dists/osminog"), Not(PathExists))
 	c.Check(filepath.Join(s.publishedStorage2.PublicPath(), "ppa/pool/contrib"), Not(PathExists))
 }
+
+func (s *PublishedRepoSuite) TestParsePrefix(c *C) {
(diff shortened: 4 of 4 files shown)


charles-dyfis-net:
The above failures look like test-suite bitrot rather than PR-specific issues to me.

charles-dyfis-net:
I don't find any preexisting tests covering the codepaths to publish _through the CLI_, so there doesn't appear to be a clear need for (or place to add) more tests than are already present in this PR as currently given. Removing from draft state.

lbolla:
My problem with this PR is that it doesn't seem possible what `storage type`, `storage name` and `paths` are for strings like:
```
storagetype:storename:path:with:colons
```
Where does `path` start? Is it `path:with:colons` or is it `with:colons` or maybe `storename:path:with:colons`?

I would just close this and document the fact that paths must not contain `:`.

@neolynx @randombenj Agree?

neolynx:
I tend to agree. if colon (:) is used as separator parsing gets messy...

what is the use case for using : in a path ?

charles-dyfis-net:
> I tend to agree. if colon (:) is used as separator parsing gets messy...
> 
> what is the use case for using : in a path ?

1. Ability to use aptly to build systems conforming with preexisting conventions that use colons.
2. As a matter of principle, avoiding adding constraints to things that do not themselves carry them (if POSIX allows all characters except NULs to be used in paths, anyone disallowing a non-NUL character is making decisions on behalf of their users that their operating system and other components do not force; leaving those decisions _to the users_ allows those users the most freedom in design).

lbolla:
@charles-dyfis-net Those are fair points, but what do you think of [my comment](https://github.com/aptly-dev/aptly/pull/959#issuecomment-1024004960)? Isn't a string like `storagetype:storename:path:with:colons` ambiguous?

charles-dyfis-net:
> @charles-dyfis-net Those are fair points, but what do you think of [my comment](https://github.com/aptly-dev/aptly/pull/959#issuecomment-1024004960)? Isn't a string like `storagetype:storename:path:with:colons` ambiguous?

Agreed that it is. One standing convention to such ambiguity (for example, used by `IFS=: read storetype storename path _` in shell) is backslash-escaping; if it were documented that `path\:with\:colons` would need to be used in the relevant locations... it's not ideal by a long shot (would also need to handle `\\` to mean a literal backslash, and arguably should bake all other backslash uses errors / values reserved for future extensions), but it avoids restricting the usable namespace.

charles-dyfis-net:
Another approach to resolving the same ambiguity is equivalent to that taken by `IFS=: read -r storetype storename path`, wherein any `:` in `path` is read as literal instead of a separator on account of `path` being the last destination; going this route, it becomes the responsibility of anyone who wants a `:` to be present in `path` to fill in `storetype` and `storename`.

These are all suboptimal, but the design decision to use a valid path character as a delimiter forced a situation where _something_ was necessarily going to be suboptimal.

neolynx:
> > what is the use case for using : in a path ?  
> 
> 1. Ability to use aptly to build systems conforming with preexisting conventions that use colons.
> 2. As a matter of principle, avoiding adding constraints to things that do not themselves carry them (if POSIX allows all characters except NULs to be used in paths, anyone disallowing a non-NUL character is making decisions on behalf of their users that their operating system and other components do not force; leaving those decisions _to the users_ allows those users the most freedom in design).

agree and agree ;-)

lbolla:
@charles-dyfis-net I think we can progress with the MR if we can do the following:

* Can we guarantee that the current behavior is unchanged by this PR, meaning, the PR is "backward compatible"?
  * This seems to be the case, as your changes make previous crashes actually work
* If so, we should take this occasion to document properly the "schema" of that string, instead of relying on the sole implementation to dictate the expected behavior.
  * Can you add it to the docs?
* Finally, rebase and merge ;-)

randombenj:
Closing this due to inactivity, feel free to leave a comment when you want to reopen it :)

Issue #973, opened 2021-08-31
Move to travis-ci.com


lbolla:
`Since June 15th, 2021, the building on travis-ci.org is ceased. Please use travis-ci.com from now on.`

We should move CI jobs to travis-ci.com instead of .org.

lbolla:
See https://docs.travis-ci.com/user/migrate/open-source-repository-migration

chhetripradeep:
Can i pick this and move CI jobs to github actions ? With recent travis-ci's security issues (https://twitter.com/peter_szilagyi/status/1437646118700175360) I don't trust them anymore. I would prefer github actions if it's okay on your end.

lbolla:
@chhetripradeep Yes of course, that'd be great!

r4co0n:
Please note that Travis CI is no longer really an option for open-source projects, as they limited their free tier, resulting in timeouts for long running jobs, see e.g. [this nextcloud/nextcloud-snap circleci jon](https://app.circleci.com/pipelines/github/nextcloud-snap/nextcloud-snap/3207/workflows/16df96ab-1aca-4989-93a8-142e11242dd7/jobs/8759). This was mitigated there by moving to GitHub Actions in nextcloud/nextcloud-snap#1939.

I would also like to see this implemented using GitHub Actions, thanks for taking this on!

PR #976, opened 2021-09-16
Add -json output flag to list|show commands

freakinhippie:
Signed-off-by: Joshua Colson <joshua.colson@gmail.com>

Fixes #

## Requirements

## Description of the Change

This PR is to add a `-json` flag to format the output of `list` and `show` commands in JSON.

Note: this initial commit is a WIP with only relevant updates to the `repo` command to solicit feedback before coding for `mirror`, `publish`, and `snapshot`.

## Checklist

- [X] documentation updated


freakinhippie:
@lbolla Would you or someone from the team be able to provide some feedback on this draft PR? I'd like to course correct sooner rather than later if there are significant problems.

Thanks.

cmd/repo_list.go
```diff
@@ -16,8 +17,10 @@ func aptlyRepoList(cmd *commander.Command, args []string) error {
 	}
 
 	raw := cmd.Flag.Lookup("raw").Value.Get().(bool)
+	jsonFlag := cmd.Flag.Lookup("json").Value.Get().(bool)
 
 	repos := make([]string, context.CollectionFactory().LocalRepoCollection().Len())
+	jsonRepos := make([]*deb.LocalRepo, context.CollectionFactory().LocalRepoCollection().Len())
```
lbolla, line 23: Are we sure that collecting full `deb.LocalRepo` objects rather than just strings (as in the `repos` array) is not going to use too much memory?

If yes, then why not *always* collect all repos in an array and do the formatting later, either with `fmt.Sprintf` or with `json.Marshal` if the `-json` flag is set?

At the moment, the same function handles both txt and json output: but it's so different, because of different variables or switches, that I wonder if it doesn't make more sense to have separate functions (if the above method doesn't work).

freakinhippie: Thanks for the review!

I'm not sure about the memory issue. Do you know if any statistics are available showing the maximum number of repo/mirror/snapshot/publish supported and/or expected?

As for the second statement, the primary reason that I chose not to do that is because I'm sure there are various scripts in the wild that parse the output of aptly and I didn't want to break user's existing scripts. I'm happy to refactor it into a separate function though, if you feel that would be more appropriate.

lbolla: I guess a good metric is to see what portion of a function are common between "txt" and "json" outputs and make a judgement if it makes sense to split them in separate methods. I'll leave this judgement to you, of course.

And sure, we should keep backward compatibility, to avoid breaking existing workflows.

freakinhippie: Hmm, I'm curious what scale of repos you have in your internal aptly deployment. I just created 10k test repos on my dev system and this was the result:

```sh
> time aptly repo list -json
[
  {
    "Name": "test-repo-00001",
    "Comment": "",
    "DefaultDistribution": "",
    "DefaultComponent": "main"
  },
 # SNIP
  {
    "Name": "test-repo-09999",
    "Comment": "Comment on repo 09999",
    "DefaultDistribution": "bionic",
    "DefaultComponent": "main"
  },
  {
    "Name": "test-repo-10000",
    "Comment": "Comment on repo 10000",
    "DefaultDistribution": "bionic",
    "DefaultComponent": "main"
  }
]

real    0m0.764s
user    0m0.258s
sys     0m0.132s

Comment (lbolla) [reply cmd/repo_list.go:23]: Actually just about 3000. I may be misremembering the slowness: I can't stop production to check again though atm.

**Comment by freakinhippie** [comment]
@lbolla I think this PR is ready for review. Thanks.

File: cmd/mirror_list.go [hunk 2]
```diff
@@ -31,7 +43,9 @@ func aptlyMirrorList(cmd *commander.Command, args []string) error {
 
 	context.CloseDatabase()
 
-	sort.Strings(repos)
+	if len(repos) > 0 {
```
lbolla, line 46: `sort.Strings` works on empty lists, too. There is no need for this "if" statement.

cmd/repo_list.go
```diff
@@ -36,13 +48,13 @@ func aptlyRepoList(cmd *commander.Command, args []string) error {
 
 	context.CloseDatabase()
 
-	sort.Strings(repos)
-
 	if raw {
+		sort.Strings(repos)
```
lbolla, line 52: These `sort.Strings(repos)` can go outside the `if raw` statement, as it was before.

cmd/repo_show.go
```diff
@@ -26,18 +32,42 @@ func aptlyRepoShow(cmd *commander.Command, args []string) error {
...
 
-	fmt.Printf("Name: %s\n", repo.Name)
-	fmt.Printf("Comment: %s\n", repo.Comment)
-	fmt.Printf("Default Distribution: %s\n", repo.DefaultDistribution)
-	fmt.Printf("Default Component: %s\n", repo.DefaultComponent)
-	if repo.Uploaders != nil {
-		fmt.Printf("Uploaders: %s\n", repo.Uploaders)
-	}
-	fmt.Printf("Number of packages: %d\n", repo.NumPackages())
+	if jsonFlag {
```
lbolla, line 35: Why is this command different from the others, in that it doesn't have two separate functions to print txt and json?

cmd/snapshot_show.go
```diff
@@ -64,14 +78,82 @@ func aptlySnapshotShow(cmd *commander.Command, args []string) error {
 		}
 	}
 
-	withPackages := context.Flags().Lookup("with-packages").Value.Get().(bool)
```
lbolla, line 67: I think it's best to leave `withPackages := ...` here instead of 40 lines above, because it's nearer to where it's first used.

deb/list.go
```diff
@@ -266,6 +266,19 @@ func (l *PackageList) Strings() []string {
 	return result
 }
 
+// FullNames builds a list of package {name}_{version}_{arch}
+func (l *PackageList) FullNames() []string {
+	result := make([]string, l.Len())
+	i := 0
+
+	for _, p := range l.packages {
```
lbolla, line 274: You don't need the `i` variable above (and the `i++` increment below). Just use the index from the `range` here.

deb/package.go
```diff
@@ -52,10 +52,10 @@ const (
 	PackageTypeInstaller = "installer"
 )
 
-// Special arhictectures
+// Special architectures
 const (
 	ArchitectureAll    = "all"
-	ArhictectureAny    = "any"
+	ArchitectureAny    = "any"
```
lbolla, line 58: Good catch!

files/README.md
```diff
@@ -12,7 +12,7 @@ This checksum is used to create a file tree where each package will reside, with
 
 ex:
 
- sha256sum **476e0**cdac6bc757dd2b78bacc1325323b09c45ecb41d4562deec2a1c7c148405 my-package_1.2.3_all.deb
+ sha256sum **476e**0cdac6bc757dd2b78bacc1325323b09c45ecb41d4562deec2a1c7c148405 my-package_1.2.3_all.deb
```
lbolla, line 15: What is this change?

lbolla (changes requested):
Nice work! I just added some minor comments.

cmd/repo_show.go
freakinhippie: I got caught up on the other commands and forgot to refactor it. I've fix that now.

deb/list.go
freakinhippie: This does need the `i`/`i++` variable because `l.packages` is type `map[string]*Package` and so the index is the map key string.

files/README.md
freakinhippie: The file tree is composed of the first _four_ characters of the checksum. The markdown was highlighting the first five.

freakinhippie:
@lbolla What is the process for aptly releases? Is there any hope for this making it into a release in the near future?

deb/list.go
lbolla: Oh, right!

lbolla (approved):
Looks good to me.

lbolla:
> @lbolla What is the process for aptly releases? Is there any hope for this making it into a release in the near future?

We haven't done a release yet since when we inherited maintenance of this project. As far as I know, releases were made with Travis, which is broken (see #973). We may figure out how to release manually in the meantime.

PR #991, opened 2021-10-26
Move to a pure Go xz implementation

lebauce:
The xz executable has to be present to uncompress control.tar.xz files causing deployment burden.
Moving to a pure Go implementation makes it a lot easier to support different distributions and platform.

### DIFF
--- a/deb/deb.go
+++ b/deb/deb.go
@@ -16,7 +16,7 @@ import (
 
 	"github.com/aptly-dev/aptly/pgp"
 	"github.com/kjk/lzma"
-	"github.com/smira/go-xz"
+	"github.com/ulikunitz/xz"
 )
 
 // Source kinds
@@ -72,7 +72,6 @@ func GetControlFileFromDeb(packageFile string) (Stanza, error) {
 				if err != nil {
 					return nil, errors.Wrapf(err, "unable to unxz %s from %s", header.Name, packageFile)
 				}
-				defer unxz.Close()
 				tarInput = unxz
 			default:
 				return nil, fmt.Errorf("unsupported tar compression in %s: %s", packageFile, header.Name)
@@ -183,7 +182,6 @@ func GetContentsFromDeb(file io.Reader, packageFile string) ([]string, error) {
 				if err != nil {
 					return nil, errors.Wrapf(err, "unable to unxz data.tar.xz from %s", packageFile)
 				}
-				defer unxz.Close()
 				tarInput = unxz
 			case "data.tar.lzma":
 				unlzma := lzma.NewReader(bufReader)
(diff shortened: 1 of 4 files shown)


lbolla:
I, too, am not convinced by the usefulness of this patch. We already rely on other commands and libraries for aptly to function, e.g. `gpg`, optionally `graphviz`, etc. Why does adding `xz` to this list makes a significant difference?

r4co0n:
@lbolla, thanks for chiming in. It should be said that command-line wrappers (like the application I built mentioned above) should be a last recourse and something calling the relevant libraries directly would be much preferred, I kinda shuddered when I saw the current implementation.

OT: My application would have solely used the API, but there were quite some things missing I really needed, resulting in much seemingly unneeded stdout-parsing (and not using the API at all to at least stay consistent in what I'm doing). This tells you how much I am not a Go developer, I might have been able to integrate something more low-level, had I simply understood what the aptly Go code actually does when I started this. btw, I think I got a green light to opensource my wrapper. Stay tuned - I will shamelessly advertise my wrapper on all relevant channels.

Still OT: I would be really happy if we could fill the missing pieces in, so I can switch to a pure API implementation. Benchmarks comparing both implementations would also be quite interesting, I think. We might also want to make output of "information commands"(show, list, …) customizable, I would suggest shamelessly stealing ideas from the good work the folks at docker aka moby have done for their [`--format option`](https://docs.docker.com/config/formatting/). Sorry for using this PR to discuss unrelated stuff, I will stop if you tell me you are annoyed about my actions.

lbolla:
I'm closing this because the pure-go xz library hasn't received updates in one year, meaning that it may be unmaintained. On the other hand, the one currently used relies on the OS `xz` which gets regular updates.

lebauce:
@lbolla @r4co0n  I understand your concern. That being said, if the system is missing xz - which may happen when you don't control the operating system where you are deploying your project that makes use of aptly, which is our case -, it will fail at the end of the process. What would you think about using the pure Go implementation as a fallback, when the `xz` utility could not found ?

randombenj:
@lebauce Fair point, however I would not go with two implementations as this might lead to different behaviours and makes maintanance difficult.
Is there a more actively maintained compression library?
@lbolla what do you think?

lbolla:
> @lbolla @r4co0n I understand your concern. That being said, if the system is missing xz - which may happen when you don't control the operating system where you are deploying your project that makes use of aptly, which is our case -, it will fail at the end of the process. What would you think about using the pure Go implementation as a fallback, when the `xz` utility could not found ?

@lebauce Understood. But if you don't control the OS where you install aptly, how can you make sure that other commands required by aptly, like `gpg` or `dot` or `bzip2`, are installed? I presume you are building aptly from source and deploying the executable on the server? Because by installing aptly via e.g. `apt`, the required dependencies are pulled in:
```
~/src/github.com/aptly-dev/aptly master ❯ apt-cache show aptly | grep -e Depends -e Suggests
Depends: bzip2, xz-utils, gnupg1, gpgv1, libc6 (>= 2.4)
Suggests: graphviz
```

How are you installing aptly and the other dependencies?

PR #1001, opened 2021-12-07
fix: Add force-replace to 'repo import' command

### LINKED ISSUE #552: Packages which only differ in file name are seen as conflicting when added to a repository
When trying to add a package to a repository (repo add <repo> <package>) which only differs in file name with one that is already in the repository, aptly reports a conflict.  
However, according to the documentation a duplicate package should only occur when packages with same tuple (package name, version, arch) and *different content* are added to the same repository.

## Context
At some point, I have to change certain characters in our package file names.  In particular I change ':' into '-' before I add packages to a FAT filesystem on an ISO.  Later on, when I add the original package to this system's repository which still has a ':' in the filename, this causes a conflict.

I tried to work around the issue by using the '-force-replace=true' flag on the repo add command.  Although this works for adding the package, it later on fails again with a conflict when trying to import a mirror from this repository into another repository.  At this point I'm stuck as the repo import command has no support for a -force-replace flag.

## Details

1. Consider a repo having the package openssh-server_1-6.6p1-2ubuntu2.8_amd64.deb
2. I rename the package to openssh-server_1:6.6p1-2ubuntu2.8_amd64.deb
3. Trying to add the package to the repository fails
```
$ sudo aptly --config packagerepo.json repo add base-repository openssh-server_1:6.6p1-2ubuntu2.8_amd64.deb
Loading packages...
[!] Unable to add package to repo openssh-server_1:6.6p1-2ubuntu2.8_amd64: conflict in package openssh-server_1:6.6p1-2ubuntu2.8_amd64
[!] Some files were skipped due to errors:
  openssh-server_1:6.6p1-2ubuntu2.8+amplidata1_amd64.deb
ERROR: some files failed to be added
```
Note that the error message does seem to indicate that the metadata of the package is twice the same.  (Actually all content is exactly the same, only the filename differs.)
4. When adding -force-replace, it works to add the package.
5. However, this fails later on when on a remote machine:
   1. Create mirror from base-repository: works
   2. Update mirror: works
   3. Import mirror to a local base-repository repository fails
```
aptly --config packagerepo.json  repo import base_mirror base-repository 'Name (~ .*)']. 
Reason: Loading packages...
ERROR: unable to import: conflict in package openssh-server_1:6.6p1-2ubuntu2.8_amd64
```
6. I don't know how to work around as there are no -force-replace flags on these commands.

Rather than providing flags in these commands, I would assume that I don't have to work around these conflicts as they shouldn't really be considered as conflicts in the first place?

fionera:
Fixes #552

## Requirements

## Description of the Change
Add support for the "--force-replace" flag to "repo import"

Why this change is important?
To allow importing repos with duplicate packages

## Checklist

- [ ] documentation updated

### DIFF
--- a/cmd/repo_import.go
+++ b/cmd/repo_import.go
@@ -23,6 +23,7 @@ Example:
 
 	cmd.Flag.Bool("dry-run", false, "don't import, just show what would be imported")
 	cmd.Flag.Bool("with-deps", false, "follow dependencies when processing package-spec")
+	cmd.Flag.Bool("force-replace", false, "when adding package that conflicts with existing package, remove existing package")
 
 	return cmd
 }
--- a/cmd/repo_move.go
+++ b/cmd/repo_move.go
@@ -74,6 +74,7 @@ func aptlyRepoMoveCopyImport(cmd *commander.Command, args []string) error {
 
 	context.Progress().Printf("Loading packages...\n")
 
+
 	dstList, err := deb.NewPackageListFromRefList(dstRepo.RefList(), context.CollectionFactory().PackageCollection(), context.Progress())
 	if err != nil {
 		return fmt.Errorf("unable to load packages: %s", err)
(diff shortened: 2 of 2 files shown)


lbolla:
Tests are missing. Please add more and rebase against master. Cheers!

randombenj:
I will close this as there seems to be a lack of interest for this change, feel free to leave a comment if you want this reopened

PR #1011, opened 2022-01-26
Trigger CI on every push

randombenj:
## Requirements

## Description of the Change

Triggers the CI build/test on every push (https://github.com/aptly-dev/aptly/pull/1010#issuecomment-1021520163) as suggested by @ximon18 

## Checklist

- [ ] ~~documentation updated~~


ximon18:
@randombenj: For some reason your "single" push triggered the workflow three times:
- [here](https://github.com/aptly-dev/aptly/actions/runs/1749551854) for commit 680e97a
- [here](https://github.com/aptly-dev/aptly/actions/runs/1749553601) for commit 4b40d77
- [here](https://github.com/aptly-dev/aptly/actions/runs/1749562475) for commit 7680b93

For the first two Git reports:
> This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository.

I don't see anything [in the documentation](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#push) that explains this. However this may be why in the projects I have worked on the trigger for the `push` event to the main branch and also trigger on the `pull_request` event.

lbolla:
FYI, I've rebased `molior-rebase` against master, so this PR will need rebase against the new `molior-rebase`, too.

randombenj:
@ximon18 That's actually intended as I force pushed 2 times :) 

I would strongly suggest going for `on: push` as not only main and PRs are tested but every single push so you immediately notice when something breaks ;)

randombenj:
@lbolla rebased the changes

ximon18 (approved):

lbolla:
Builds for go 1.13 fail because of the "azure" library, which is using features from go 1.14+: https://pkg.go.dev/net/http#Header.Values

lbolla:
1.17 fails. I've opened a separate ticket: https://github.com/aptly-dev/aptly/issues/1012

randombenj:
@lbolla shall we reenable 1.17?

lbolla (approved):

PR #1021, opened 2022-02-05
Fix the installer path for Ubuntu Focal

### LINKED ISSUE #1018: Installer location for Ubuntu focal has changed.
Ubuntu Focal has changed the installer to the `live` installer and has moved the location of the installer from
https://atl.mirrors.clouvider.net/ubuntu/dists/focal/main/installer-amd64/current/**images**/SHA256SUMS
to
https://atl.mirrors.clouvider.net/ubuntu/dists/focal/main/installer-amd64/current/**legacy-images**/SHA256SUMS

## Detailed Description
Aptly should know that focal and possibly newer versions have a different path for the install images.

## Context
We can't cache and snapshot installer images for focal.

## Possible Implementation
I'm going to see if I can create a patch and share it. I haven't dug into the code yet.

## Your Environment
We are mirroring 116 mirrors and consolidating them down to three for three flavors of Ubuntu (nine total). We keep snapshot indefinitely and rotate them with `tags` (symlinks) of common names to the actual snapshot. We are adding critical patches to the snapshot and republishing as needed.

rldleblanc:
Ubuntu has started depreciating the Debian installer in focal
and moved the installer images to a different path. In versions
after focal, they are completely removed. This basically gives
us more time to figure out how to use the new system.

Fixes #1018

## Requirements

This is the first time I've worked with Go, so I'm not familiar with the test suite. I needed this quick and am contributing back my fixes. I've tested that it properly mirrors the new `legacy-images` properly. You are welcome to ignore it if you would rather do things differently. I just don't have any additional time to contribute.

## Checklist

- [ ] documentation updated

### DIFF
--- a/deb/index_files.go
+++ b/deb/index_files.go
@@ -242,7 +242,7 @@ func newIndexFiles(publishedStorage aptly.PublishedStorage, basePath, tempDir, s
 	}
 }
 
-func (files *indexFiles) PackageIndex(component, arch string, udeb, installer bool) *indexFile {
+func (files *indexFiles) PackageIndex(component, arch string, udeb bool, installer bool, distribution string) *indexFile {
 	if arch == ArchitectureSource {
 		udeb = false
 	}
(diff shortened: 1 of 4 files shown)


NeroBurner:
is this change necessary starting with Ubuntu 20.04 focal? Then we'll need those changes for newer Ubuntu versions as well. To not encode each and every Ubuntu release name we could compare with the release number (for example `release >= 20.04`)

rldleblanc:
Well, Ubuntu has really made a mess of this. 20.04 is the only one (still supported) that has the legacy installer. In 22.04 it won't exist at all. I couldn't get the 20.04 installer to work, so I'm just biting the bullet and moving over to their new install method as it's their only "supported" method going forward. If this doesn't get merged I won't be heartbroken.

NeroBurner:
I don't understand what the installer path is and how it affects the aptly repository. Would you be so kind to point me to some kind of documentation or explain it to me please?

NeroBurner:
thank you for this thorough explanation :bow:

j1mc:
Is the suggested patch here not likely to be merged, then? I know for certain that the 22.04 legacy installer images are not present at all.

neolynx:
See https://github.com/aptly-dev/aptly/pull/1248 (rebased on master).

Debian bookworm and Ubuntu focal seem to work :+1: 

However, Ubuntu/jammy does not seem to work, no installer files are downloaded, which are on the mirror:
http://archive.ubuntu.com/ubuntu/dists/jammy/main/

Does anyone know about jammy installers ?

neolynx:
fixed in https://github.com/aptly-dev/aptly/pull/1248

PR #961, opened 2021-06-09
feat: Add etcd database support

hudeng-go:
improve concurrent access and high availability of aptly with the help of the characteristics of etcd

## Requirements

All new code should be covered with tests, documentation should be updated. CI should pass.

## Checklist

- [x] unit-test added (if change is algorithm)
- [x] functional test added/updated (if change is functional)
- [x] man page updated (if applicable)
- [x] bash completion updated (if applicable)
- [x] documentation updated
- [x] author name in `AUTHORS`

PR #1022, opened 2022-02-07
feat: Add etcd database support

hudeng-go:
improve concurrent access and high availability of aptly with the help of the characteristics of etcd

Fixes #

## Requirements

## Checklist

- [x] documentation updated

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,7 +40,7 @@ Please report unacceptable behavior to [team@aptly.info](mailto:team@aptly.info)
 
 There are two kinds of documentation:
 
-* [aptly website](https://www.aptly/info)
+* [aptly website](https://www.aptly.info)
 * aptly `man` page
 
 Core content is mostly the same, but website contains more information, tutorials, examples.


hudeng-go:
@lbolla I created a new PR to solve the previous problem at #961.

Makefile
```diff
@@ -47,7 +60,7 @@ ifeq ($(RUN_LONG_TESTS), yes)
 	PATH=$(BINPATH)/:$(PATH) && . system/env/bin/activate && APTLY_VERSION=$(VERSION) $(PYTHON) system/run.py --long $(TESTS)
 endif
 
-test:
+test: etcd-prepare
```
lbolla, line 63: This will re-download `etcd` everytime tests run, right? That seems quite wasteful. Maybe we can check if etcd exists and don't redownload it every time?

utils/config.go
```diff
@@ -33,6 +33,7 @@ type ConfigStructure struct { // nolint: maligned
 	SwiftPublishRoots      map[string]SwiftPublishRoot      `json:"SwiftPublishEndpoints"`
 	AzurePublishRoots      map[string]AzurePublishRoot      `json:"AzurePublishEndpoints"`
 	AsyncAPI               bool                             `json:"AsyncAPI"`
+	DatabaseEtcd           string                           `json:"databaseEtcd"`
```
lbolla, line 36: I think you also need to add a default value to line ~100 below. See failing tests.

randombenj:
@lbolla  @hudeng-go as this is quite a 'big' change in the sense that it does affect the central database, what do you think ab out running the system tests both with leveldb and etcd? Or make them configurable and we can run them in the pipeline in paralell?

PR #1023, opened 2022-02-08
feat: Add etcd database support

hudeng-go:
improve concurrent access and high availability of aptly with the help of the characteristics of etcd

Fixes #

## Requirements

## Checklist

- [x] documentation updated

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,13 +23,10 @@ help:  ## Print this help
 	@grep -E '^[a-zA-Z][a-zA-Z0-9_-]*:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 
 prepare:
-<<<<<<< HEAD
 	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(shell go env GOPATH)/bin v1.51.2
 
 etcd-prepare:
-=======
 	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(shell go env GOPATH)/bin v1.43.0
->>>>>>> d03d2325 (feat: Add system test for etcd)
 	# etcd test prepare
 	rm -rf /tmp/etcd-download-test/test-data && mkdir -p /tmp/etcd-download-test/test-data
 	if [ ! -e /tmp/etcd-${ETCD_VER}-linux-amd64.tar.gz ]; then curl -L ${DOWNLOAD_URL}/${ETCD_VER}/etcd-${ETCD_VER}-linux-amd64.tar.gz -o /tmp/etcd-${ETCD_VER}-linux-amd64.tar.gz; fi
@@ -83,8 +80,8 @@ ifeq ($(RUN_LONG_TESTS), yes)
 	if [ ! -e ~/aptly-fixture-db ]; then git clone https://github.com/aptly-dev/aptly-fixture-db.git ~/aptly-fixture-db/; fi
 	if [ ! -e ~/aptly-fixture-pool ]; then git clone https://github.com/aptly-dev/aptly-fixture-pool.git ~/aptly-fixture-pool/; fi
 	# TODO: maybe we can skip imgrading levledb data to etcd
-	PATH=$(BINPATH)/:$(PATH) && . system/env/bin/activate && APTLY_VERSION=$(VERSION) $(PYTHON) system/leveldb2etcd.py --datadir ~/aptly-fixture-db
-	PATH=$(BINPATH)/:$(PATH) && . system/env/bin/activate && APTLY_ETCD_DATABASE="127.0.0.1:2379" APTLY_VERSION=$(VERSION) $(PYTHON) system/run.py --long $(TESTS)
+	PATH=$(BINPATH)/:$(PATH) && . system/env/bin/activate && $(PYTHON) system/leveldb2etcd.py --datadir ~/aptly-fixture-db
+	PATH=$(BINPATH)/:$(PATH) && . system/env/bin/activate && APTLY_DATABASE_TYPE="etcd" APTLY_DATABASE_URL="127.0.0.1:2379" APTLY_VERSION=$(VERSION) $(PYTHON) system/run.py --long $(TESTS)
 endif
 
 bench:
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ COVERAGE_DIR?=$(shell mktemp -d)
 ETCD_VER=v3.5.2
 DOWNLOAD_URL=https://storage.googleapis.com/etcd
 
-all: modules test bench check system-test
+all: modules test bench check system-test system-test-etcd
 
 # Self-documenting Makefile
 # https://marmelab.com/blog/2016/02/29/auto-documented-makefile.html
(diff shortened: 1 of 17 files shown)


hudeng-go:
@lbolla Resolve system test errors at #1022

lbolla:
@hudeng-go Looking good to me. What do you think of this comment on the other ticket? https://github.com/aptly-dev/aptly/pull/1022#issuecomment-1031567248

system/run-etcd.py
```diff
@@ -0,0 +1,175 @@
+#!/usr/bin/env python
```
lbolla, line 1: This script is almost identical to `system/run.py`:
```
/tmp/aptly etcd-dev ❯ diff system/run.py system/run-etcd.py
97a98,100
>                     t.configOverride = {
>                         "databaseEtcd": "127.0.0.1:2379"
>                     }
```
It would be much better to have a flag or an env variable to tell the test runner script to use etcd, rather than duplicate the whole script.

Moreover, I suspect that using `configOverride` may not work on all cases. Many unittests set their own overrides, so either they will be overwritten with the `databaseEtcd` key or viceversa.

Maybe `databaseEtcd` could come from the environment, e.g. `APTLY_ETCD_DATABASE` and set in `lib.py:BaseTest.config` dictionary?

lbolla (changes requested):

system/lib.py
```diff
@@ -142,6 +142,7 @@ class BaseTest(object):
         "gpgDisableSign": False,
         "ppaDistributorID": "ubuntu",
         "ppaCodename": "",
+        "databaseEtcd": "%s" % os.environ["APTLY_ETCD_DATABASE"],
```
lbolla, line 145: This will throw a `KeyError` if the env var is not set, which it isn't in the normal `system-tests`. Maybe use `os.environ.get("APTLY_ETCD_DATABASE") or ""` to make sure we have a valid value in the config?

context/context.go
```diff
@@ -287,7 +288,12 @@ func (context *AptlyContext) _database() (database.Storage, error) {
 	if context.database == nil {
 		var err error
 
-		context.database, err = goleveldb.NewDB(context.dbPath())
+		if context.config().DatabaseEtcd != "" {
```
lbolla, line 291: I wonder if we should have a more explicit way to pick the database, rather than look at `DatabaseEtcd`.

Maybe having a new config like `databaseBackend` to be one of: "leveldb", "etcd", etc.? What do you think?

lbolla:
@hudeng-go Tests failed... https://github.com/aptly-dev/aptly/runs/5140408342?check_suite_focus=true
They also log *a lot* and take quite a long time, which will eat our free allowance of GH actions... Can you think of a solution?

hudeng-go:
> @hudeng-go Tests failed... https://github.com/aptly-dev/aptly/runs/5140408342?check_suite_focus=true They also log _a lot_ and take quite a long time, which will eat our free allowance of GH actions... Can you think of a solution?

Sorry, it took a long time to link etcd when migrating data. We are looking for the reason.

lbolla:
@hudeng-go I had to stop the CI once again, as it was stuck... Can we make the script so that it would crash at the first error, so we don't continue a clearly failing test?

hudeng-go:
> @hudeng-go I had to stop the CI once again, as it was stuck... Can we make the script so that it would crash at the first error, so we don't continue a clearly failing test?

I have modified the migration script to exit with an error. Sorry again.

hudeng-go:
> 



> @hudeng-go I had to stop the CI once again, as it was stuck... Can we make the script so that it would crash at the first error, so we don't continue a clearly failing test?

In my environment, data can be migrated to normally. I don't know why I can't connect to etcd database in GH action environment.

Makefile
```diff
@@ -47,6 +56,15 @@ ifeq ($(RUN_LONG_TESTS), yes)
 	PATH=$(BINPATH)/:$(PATH) && . system/env/bin/activate && APTLY_VERSION=$(VERSION) $(PYTHON) system/run.py --long $(TESTS)
 endif
 
+system-test-etcd: install system/env
+ifeq ($(RUN_LONG_TESTS), yes)
+	if [ ! -e ~/aptly-fixture-db ]; then git clone https://github.com/aptly-dev/aptly-fixture-db.git ~/aptly-fixture-db/; fi
+	if [ ! -e ~/aptly-fixture-pool ]; then git clone https://github.com/aptly-dev/aptly-fixture-pool.git ~/aptly-fixture-pool/; fi
+	# TODO: maybe we can skip imgrading levledb data to etcd
+	PATH=$(BINPATH)/:$(PATH) && . system/env/bin/activate && $(PYTHON) system/leveldb2etcd.py --datadir ~/aptly-fixture-db
+	PATH=$(BINPATH)/:$(PATH) && . system/env/bin/activate && APTLY_DATABASE_TYPE="etcd" APTLY_DATABASE_TYPE="127.0.0.1:2379" APTLY_VERSION=$(VERSION) $(PYTHON) system/run.py --long $(TESTS)
```
lbolla, line 65: I guess the second APTLY_DATABASE_TYPE should read APTLY_DATABASE_URL instead?

context/context.go
```diff
@@ -287,7 +288,12 @@ func (context *AptlyContext) _database() (database.Storage, error) {
 	if context.database == nil {
 		var err error
 
-		context.database, err = goleveldb.NewDB(context.dbPath())
+		if context.config().DatabaseBackend.Type == "etcd" {
+			context.database, err = etcddb.NewDB(context.config().DatabaseBackend.Url)
+		} else {
+			context.database, err = goleveldb.NewDB(context.dbPath())
```
lbolla, line 294: I see the reason for using `DatabaseBackend.Type` as recommended. I wonder though if we can adapt `dbPath` (i.e. `rootPath/db`) into your new `DbConfig` structure? We must be backward compatible though. Maybe it's worth raising a separate ticket once this is merged.

lbolla:
> @hudeng-go I had to stop the CI once again, as it was stuck... Can we make the script so that it would crash at the first error, so we don't continue a clearly failing test?

Once https://github.com/aptly-dev/aptly/pull/1033 is merged, you can rebase to ensure we have stricter timeouts for CI jobs, and avoid tests wasting our GHA allowance.

lbolla:
> > @hudeng-go I had to stop the CI once again, as it was stuck... Can we make the script so that it would crash at the first error, so we don't continue a clearly failing test?
> 
> Once #1033 is merged, you can rebase to ensure we have stricter timeouts for CI jobs, and avoid tests wasting our GHA allowance.

#1033 is merged. Please rebase this branch before carrying out further tests. Thanks!

lbolla:
@hudeng-go On my humble LXD image, which I use for development, `etcd` is using quite a lot of resources. Maybe it crashes on CI because it's too memory/cpu hungry?
Screenshot from 2022-02-14 14-09-04

Makefile
hudeng-go: Yes, my mistake.

hudeng-go:
> @hudeng-go On my humble LXD image, which I use for development, `etcd` is using quite a lot of resources. Maybe it crashes on CI because it's too memory/cpu hungry? Screenshot from 2022-02-14 14-09-04

Is aptly in makefile_ DATABASE_ It's my fault that the URL environment variable is incorrectly configured

context/context.go
hudeng-go: I don't quite understand. At present, this dbpath should be available only for leveldb, and the current processing method should be compatible with it. My idea is that if the dbconfig configuration in the configuration file is empty, the default leveldb is used as the back-end, and leveldb compatibility should not be required in dbconfig. This is also explained in the man document.

tomsiewert:
Is there any update to bring this patch into upstream?

tomsiewert:
@randombenj As you seem to be the most active reviewer/contributor for Aptly, could you please review this?

neolynx:
Hi all,

maybe we should rebase this on master first, which seems to be a bit of work.. any chance someone could help with this ?

neolynx:
Hi all,

could anyone have a look at why the test is running into a timeout in https://github.com/aptly-dev/aptly/pull/1284 ?

neolynx:
please git the maintainers force push permissions, so the branch on the PR git repo can be rebased.

In the meanwhile, see: https://github.com/aptly-dev/aptly/pull/1304

neolynx:
Hi all,

it's been a while, but I guess 2 years late is better then never. Please test https://github.com/aptly-dev/aptly/pull/1304 as it is ready to merge soon !

PR #1028, opened 2022-02-10
[WORK IN PROGRESS] Initial commits to bring ldap authentication and repo ldap group permissions

bpiraeus:
## Requirements
github.com/gin-gonic/contrib/sessions
github.com/gin-gonic/contrib/cookies
github.com/go-ldap/ldap/v3

INCOMPLETE 

## Description of the Change
This brings LDAP authentication to the API to protect it from unwanted intrusions, it also adds ldap group permissions on a per repo basis.

## Checklist

- [ ] documentation updated


bpiraeus:
Feedback, nits, gnats, and other fine offerings are accepted, there's still much work to be done, but I'd rather get meaningful feedback before I go too far down any given rabbit hole.

api/ldap.go
```diff
@@ -0,0 +1,101 @@
...
+	// connect to ldap server
+	server := fmt.Sprintf("%s", config.LdapServer)
+	conn, err := ldap.Dial("tcp", server)
+	if err != nil {
+		return err
+	}
+	defer conn.Close()
+
+	// reconnect via tls
+	err = conn.StartTLS(&tls.Config{InsecureSkipVerify: true})
```
lbolla, line 31: The choice to use "insecure" should probably be configurable.

api/repos.go
```diff
@@ -81,6 +84,10 @@ func apiReposEdit(c *gin.Context) {
 		return
 	}
 
+	if !checkGroup(repo.LdapGroup) {
```
lbolla, line 87: Where is `checkGroup` defined? I can see `CheckGroup` (capital "C") only.

api/repos.go
```diff
@@ -38,6 +38,7 @@ func apiReposCreate(c *gin.Context) {
 		Comment             string
 		DefaultDistribution string
 		DefaultComponent    string
+		LdapGroup           string
```
lbolla, line 41: IIUC, if `LdapGroup` is an empty string, it means "no ldap used in this repo". Maybe it would be best to use `nil` for that purpose?

api/repos.go
```diff
@@ -139,6 +149,10 @@ func apiReposDrop(c *gin.Context) {
 		return
 	}
 
+	if !checkGroup(repo.LdapGroup) {
```
lbolla, line 152: Given that this "if" check is repeated identical multiple times, I would refactor it.

api/router.go
```diff
@@ -48,96 +54,115 @@ func Router(c *ctx.AptlyContext) http.Handler {
...
-		root.PUT("/publish/:prefix/:distribution", apiPublishUpdateSwitch)
-		root.DELETE("/publish/:prefix/:distribution", apiPublishDrop)
-	}
+	router.GET("/version", apiVersion)
 
+	var username string
+	var password string
+	router.POST("/login", func(c *gin.Context) {
+		session := sessions.Default(c)
+		if config.UseAuth {
```
lbolla, line 75: `UseAuth` will have to be documented at some point.

api/router.go
```diff
@@ -48,96 +54,115 @@ func Router(c *ctx.AptlyContext) http.Handler {
...
+			log.Printf("UseAuth is enabled\n")
+			username = c.PostForm("username")
+			password = c.PostForm("password")
+			err := Authorize(username, password)
+			if err != nil {
+				c.AbortWithError(403, err)
+			}
+			log.Printf("%s authorized from %s\n", username, c.ClientIP())
+		}
+		session.Set(token.String(), time.Now().Unix())
```
lbolla, line 85: How long does the session last? Is it possible to wipe it, e.g. via a `logout` method?

utils/config.go
```diff
@@ -86,6 +90,10 @@ type AzurePublishRoot struct {
 // Config is configuration for aptly, shared by all modules
 var Config = ConfigStructure{
 	RootDir:                filepath.Join(os.Getenv("HOME"), ".aptly"),
+	UseAuth:                false,
+	LdapServer:             "127.0.0.1:389",
```
lbolla, line 94: I don't think we want these settings to be the default ones. By default, I guess we want them emptly, i.e. no ldap configured.

lbolla (commented):
Overall, it's an interesting feature to have, IMO.

I added some comments specific to the code. More generally, I think if we should consider implementing a more generic "auth" layer, of which "ldap" is just an implementation: there may be other ways to implement it, too. E.g. maybe it's worth considering to have a `type Auth interface` which wraps the required functionality and a `struct LdapAuth` that implements it. We can also have a `struct NoAuth` which is used when `config.UseAuth` is false, to keep things consistent.

Just food for thought, anyway. I'd like to hear other opinions, too. @randombenj @neolynx ?

api/router.go
bpiraeus: added a logout method, and default timeout is 1 hour

utils/config.go
bpiraeus: cleared them up and redesigned some of it

api/router.go
bpiraeus: agreed *sigh* I'm terrible at docs

api/repos.go
bpiraeus: has been :)

api/repos.go
bpiraeus: redesigned quite a bit to manage this cleaner

api/ldap.go
bpiraeus: agreed

api/repos.go
bpiraeus: cleaned all of that up

lbolla:
@bpiraeus I've run the test suite and `staticcheck` complains of some unsupported packages. https://github.com/aptly-dev/aptly/runs/5356669809?check_suite_focus=true

bpiraeus:
> @bpiraeus I've run the test suite and `staticcheck` complains of some unsupported packages. https://github.com/aptly-dev/aptly/runs/5356669809?check_suite_focus=true

Agreed, I believe those are coming from the ldap library, but I've not had opportunity to track them down as I've been crazy busy with work this recent week. On the up side, I feel like I've made good progress with re-implementing it with a more generic handler.

PR #1033, opened 2022-02-12
Timeout CI build job after 30 minutes

### LINKED ISSUE #1032: Consider adding timeouts to CI jobs to avoid runaway tests
## Detailed Description

See https://github.com/aptly-dev/aptly/pull/1023#issuecomment-1035983552

In order to avoid running out of our limited quota of GH Actions, I think it would make sense to add a timeout to CI jobs.
I don't know if GH Actions supports it.

lbolla:
Fix #1032

Fixes #

## Description of the Change

Add a 30 minutes timeout to CI build test, to avoid runaway tests.
Currently sane unittests run in less than 10 minutes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,7 @@ jobs:
     name: test
     runs-on: ubuntu-20.04
     continue-on-error: ${{ matrix.allow_failure }}
+    timeout-minutes: 30
     strategy:
       fail-fast: true
       matrix:


ximon18:
Looks good. The only thing I don't know is what happens if one of the matrix jobs has to wait for a free hosted runner before it can start - does the 30 minutes apply from the start of each matrix job or is across all matrix jobs and thus ones that started late could get timed out earlier than 30 minutes? I read the docs but didn't find the answer to this. I wondered because if you say sane jobs finish in 10 minutes then maybe 30 minutes is a bit on the high side and could be lower.

ximon18 (approved):

lbolla:
Failing test is caused by linux.del.com being down, which seems usual: https://www.dell.com/community/Systems-Management-General/linux-dell-com-down-or-decomissioned/td-p/7731548

randombenj (approved):
LGTM!

PR #1038, opened 2022-02-24
Fix/S3 too many object listing

daikw:
## Requirements

- `aptly publish` becomes too slow to execute, if you use S3 as your storage, and your S3 bucket has full of objects.
- This is because `s3.ListObjectsPages` (used in `PublishedStorage#internalFilelist`) without `Prefix` option scans all of object inside the specified bucket.

## Description of the Change

- calculate `prefix` value from `publishedDirectory`
- call `PublishedStorage#internalFilelist` with `prefix` value

## Checklist

- ~[ ] unit-test added (if change is algorithm)~
- ~[ ] functional test added/updated (if change is functional)~
- ~[ ] man page updated (if applicable)~
- ~[ ] bash completion updated (if applicable)~
- ~[ ] documentation updated~

### DIFF
--- a/s3/public.go
+++ b/s3/public.go
@@ -297,11 +297,18 @@ func (storage *PublishedStorage) RemoveDirs(path string, progress aptly.Progress
 func (storage *PublishedStorage) LinkFromPool(publishedDirectory, fileName string, sourcePool aptly.PackagePool,
 	sourcePath string, sourceChecksums utils.ChecksumInfo, force bool) error {
 
+	// in: prefix/pool/component/liba/libav/
+	// out: prefix
+	directoryRoot := func(path string) string {
+		return strings.Split(path, "/")[0]
+	}
+
+	prefix := directoryRoot(publishedDirectory)
 	relPath := filepath.Join(publishedDirectory, fileName)
 	poolPath := filepath.Join(storage.prefix, relPath)
 
 	if storage.pathCache == nil {
-		paths, md5s, err := storage.internalFilelist("", true)
+		paths, md5s, err := storage.internalFilelist(prefix, true)
 		if err != nil {
 			return errors.Wrap(err, "error caching paths under prefix")
 		}
--- a/AUTHORS
+++ b/AUTHORS
@@ -44,3 +44,4 @@ List of contributors, in chronological order:
 * Andre Roth (https://github.com/neolynx)
 * Lorenzo Bolla (https://github.com/lbolla)
 * Benj Fassbind (https://github.com/randombenj)
+* Daiki Watanabe (https://github.com/daikw)


s3/public.go
```diff
@@ -297,11 +297,18 @@ func (storage *PublishedStorage) RemoveDirs(path string, progress aptly.Progress
...
 
+	// in: prefix/pool/component/liba/libav/
+	// out: prefix
+	directoryRoot := func(path string) string {
+		return strings.Split(path, "/")[0]
+	}
+
+	prefix := directoryRoot(publishedDirectory)
 	relPath := filepath.Join(publishedDirectory, fileName)
 	poolPath := filepath.Join(storage.prefix, relPath)
```
r4co0n, line 308: I think `storage.prefix` already contains the `directoryRoot` you introduced. This will also use the full prefix if it containes `/`.

randombenj:
Closing this due to inactivity, feel free to leave a comment when you want to reopen it :)

s3/public.go
daikw: Sorry for my extremely late feedback.
I completely have forgotten this PR after writing, but I have chance to revise it now.

In conclusion, in our project, aptly config `S3PublishEndpoints.[].prefix` is not precisely configured.
This PR's change is completely problematic. I'm grad to notice this. Thank you!

daikw: Sorry for my extremely late feedback.
I completely have forgotten this PR after writing, but I have chance to revise it now.

---

> `storage.prefix` already contains the `prefix` you introduced

According to following survey, I think this is not true.

`storage.prefix` is like `test.s3.bucket.com`, and `prefix` I introduced is like `published_directory_root`.

But this PR is completely problematic (test fails, other storage implementations are ignored, and so on), so I'd like to think another patch later. Thanks in advance!

---

There are some `prefix`es in the aptly project. I think at least 2 prefixes are used.

- [`PublishedStorage#prefix`](https://github.com/aptly-dev/aptly/blob/aeef41bf70de25bdbe8ef456328d24c989b909df/s3/public.go#L50)
    - The implementation of the interface [`PublishedStorage`](https://github.com/aptly-dev/aptly/blob/aeef41bf70de25bdbe8ef456328d24c989b909df/aptly/interfaces.go#L63)
    - This is configured by [`XxxPublishEndpoints.[].prefix`](https://github.com/aptly-dev/aptly/blob/aeef41bf70de25bdbe8ef456328d24c989b909df/utils/config.go#L58)
- [`PublishedRepo#Prefix`](https://github.com/aptly-dev/aptly/blob/aeef41bf70de25bdbe8ef456328d24c989b909df/deb/publish.go#L39)
    - `Package` does not have `prefix` field. Instead, some `Package`'s method have `prefix` argument.

storage's prefix does not equal to repo's prefix.

---

In this PR, I'd like to restrict `PublishedStorage#internalFilelist`'s scanning scope by using a prefix like `test.s3.bucket.com/published_directory_root`, instead of `test.s3.bucket.com/`.

When we exec `aptly publish snapshot` command, method call is like following:

1. [`PublishedRepo#Publish`](https://github.com/aptly-dev/aptly/blob/aeef41bf70de25bdbe8ef456328d24c989b909df/deb/publish.go#L553)
1. [`Package#LinkFromPool`](https://github.com/aptly-dev/aptly/blob/aeef41bf70de25bdbe8ef456328d24c989b909df/deb/package.go#L615)
    - `publishedDirectory` is calculated from `prefix` argument (which is `PublishedRepo#Prefix`)
1. [`PublishedStorage#LinkFromPool`](https://github.com/aptly-dev/aptly/blob/aeef41bf70de25bdbe8ef456328d24c989b909df/s3/public.go#L305)
1. [`PublishedStorage#internalFilelist`](https://github.com/aptly-dev/aptly/blob/aeef41bf70de25bdbe8ef456328d24c989b909df/s3/public.go#L375C46-L375C46)
    - `storage.prefix` is used as the base prefix (which is `PublishedStorage#prefix`)

repo's prefix information is included in `publishedDirectory`, but it is not used in scanning files.
That is why I used `prefix := directoryRoot(publishedDirectory)`.

PR #1042, opened 2022-03-13
Adding authorisation options for API access

bpiraeus:
Fixes #

## Requirements

## Checklist

- [ ] documentation updated

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -63,3 +63,4 @@ List of contributors, in chronological order:
 * Ramón N.Rodriguez (https://github.com/runitonmetal)
 * Golf Hu (https://github.com/hudeng-go)
 * Cookie Fei (https://github.com/wuhuang26)
+* Brett Hawn (https://github.com/bpiraeus)
--- a/api/auth.go
+++ b/api/auth.go
@@ -1,0 +1,119 @@
+package api
+
+import (
+	"crypto/tls"
+	"fmt"
+	"strings"
+
+	"github.com/gin-contrib/sessions"
+	"github.com/gin-gonic/gin"
+	"github.com/go-ldap/ldap/v3"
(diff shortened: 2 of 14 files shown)


bpiraeus:
I went through the failing tests, and unless I'm missing something, those are not from my changes, not sure what I can do about that short of refactoring someone else's code. I'm happy to muck around if needed, but you'd likely prefer someone with a lot more chops than myself to mitigate those issues.

randombenj:
@bpiraeus It looks like mostly deprecation errors 

- https://github.com/golang/go/issues/44226
- package golang.org/x/crypto/ssh/terminal is deprecated: this package moved to golang.org/x/term

If you want to, feel free to investigate if there are some simple fixes for the deprecated packages (x/term) seems to be quite an easy one, not sure about the crypto one ...

bpiraeus:
Cleaned up a few oddballs in my code that gosimple flagged, also moved the terminal bits to the appropriate package, the pgp ones appear as if they're likely going to want 3rd party library.

randombenj:
@bpiraeus I just re reviewd the changes in `go.mod` and it seems like you upgraded quite a lot of dependencies. Was this on purpose? I think if you just add the packages you need to add without upgrading the others it will 'fix' the linting.

I would propose to upgrade all the other deps in a sepparate change. What do you think?

neolynx:
introducing authorization per repo is a nice idea. 

however, authentication without SSL is not that nice, so probably we should implement this as well with this change. 

since different authentication types are supported, the variable `LdapGroup` should probably `AuthorizedGroup`

neolynx:
rebased on master

neolynx:
@bpiraeus are you still interested in continuing this ?

Issue #920, opened 2020-04-30
Is Aptly still maintained?


NielsH:
Hi,

We're considering switching to Aptly, however we are having some doubts because it seems somewhat abandoned; the last commits stop around october 2019 and when I look at the website regarding the Roadmap I see planned releases for Dec 2019 and Jun 2020 which appear to be out-of-date ( https://www.aptly.info/download/ ) since it has already been Dec 2019 of course.

Is it still maintained or no more?

Cheers

sjquinney:
I've recently been wondering the same thing. The website is in a poor state with the tutorials having lost all the examples which makes it very hard to get started. Also questions on the mailing list go unanswered. Would be a shame as aptly looks really good compared to other tools which I've found lacking in features.

adam-kosseck:
Take a look at the developer's Github activity, he is still very active just not on this project.
It seems he is working at a Kubernetes startup and doesn't have time for this project anymore. 

What are the best alternatives to Aptly?  Well I've tried both the aptmirror and debmirror CLI tools, at the time debmirror seemed more mature but didn't have the delta or lifecycle support that Aptly has.  In the UI space I've tried both Foreman (katello) and its commercial (.deb focussed) offshoot Orcharhino, both seemed promising and had all of the required features but had trouble working reliably in our admittedly highly complex environments.

My next point of investigation was going to be Pulp3, as this is the CLI tool that Foreman/Katello & Orcharhino use under the hood...

berendt:
@adam-kosseck We have also used Aptly so far and will now look at Pulp3 as we also need mirrors for Docker/Podman images and Ansible roles. I think Pulp3 is a good solution then. Aptly is definitely no longer an option for us.

cgallay:
Hi, @adam-kosseck which alternative did you end up using?

adam-kosseck:
I'm still using Aptly due to the amount of time I've already invested in wrapper scripts to automate processes across multiple environments.
However I remain keen to investigate Pulp 3 when time permits.

seitzbg:
Any comments from either of the maintainers regarding Aptly?  Would love to start using it but not if it is a dead project.  Pulp3 looks interesting but no CLI right now (which isn't a deal breaker but that's why I want to use aptly).

berendt:
@seitzbg Have a look at https://github.com/pulp/pulp_ansible. This makes it easy to manage.

oberones:
[...]

PR #1055, opened 2022-04-04
Publish releases and nightly builds from ci

randombenj:
## Description of the Change

#920 

We moved the whole aptly infrastructure to our own infrastructure and moved `internal.aptly.info` to `aptly-ops.aptly.info`
so we can publish new releases there. :tada: 

## Checklist

- [ ] documentation updated

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,6 +102,22 @@ jobs:
         run: |
           make release
 
+      - name: Publish nightly release to aptly
+        if: github.ref == 'refs/heads/master'
+        env:
+          APTLY_USER: ${{ secrets.APTLY_USER }}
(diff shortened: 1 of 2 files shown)


ximon18 (approved):
I'm not familiar with the infrastructure that will host the release or how that works but the changes in this PR seen fairly minimal and self-evident.

The only concern I can think of is that it seems that the same credentials are used for publishing both master nightly and tagged release artefacts and if there is somewhere in the nightly setup a weakness through which the credentials could be obtained then the damage would not be limited to nightly builds but could also be used to compromise actual releases.

randombenj:
@ximon18 Thanks for the review. Yeah thats a fair point, however, I will keep the setup as is for now as it was like this before as well. I think this can still be improved later on.

PR #1063, opened 2022-04-22
Support custom Azure publish endpoint

chuan:
## Requirements

## Description of the Change

Add custom endpoint config for publishing to Azure blob storage. This is needed to support publishing to Azure sovereign clouds, i.e. China, Germany, and the US government. The change also made it possible to test the Azure publishing via [the Azure storage emulator](https://github.com/Azure/Azurite) by setting the endpoint to the local emulator address.

## Checklist

- ~~[ ] functional test added/updated (if change is functional)~~
- ~~[ ] bash completion updated (if applicable)~~
- [x] documentation updated


randombenj (approved):
LGTM

Issue #428, opened 2016-08-01
can't recover from snapshot not found


damaex17:
Error #01: snapshot with uuid 68358b45-7aba-4c7a-9280-8a1c55c4c20d not found 
     Meta: Operation aborted

This error shows up whenever I try to publish.

damaex17:
# aptly --config=/home/aptly/.aptly.conf publish list
ERROR: unable to load list of repos: snapshot with uuid 68358b45-7aba-4c7a-9280-8a1c55c4c20d not found

damaex17:
i had to drop the published distribution and re-upload it. i couldn't find another way around it...

please close if this is intended

scravy:
running into the same issue here.

How do I drop a whole distribution and re-upload it?

scravy:
solved it by just dropping the local database from the aptly host. the stuff in s3 continued to work.

smira:
hmm... any details on the issue? it should never happen unless there's some bug out there

smutel:
Currently we have the same issue. 
```sh
$ aptly publish list
ERROR: unable to load list of repos: snapshot with uuid cfe62615-165a-4603-8f58-886cb2bd2b41 not found
```
As the snapshot is no longer there , how to find the publish to remove ?
Anyway to find the publish in error directly in the db ?

I tried also a aptly db recover / aptly db cleanup without any change.

neolynx:
most likely fixed since then

PR #1064, opened 2022-04-22
Give info when unable to load list of repos

smutel:
This PR is a workaround for #428.

## Requirements

## Description of the Change

The idea is to have information about publish which is in trouble and to be able to drop it and to re-publish it.

## Checklist

- [ ] documentation updated


randombenj:
Thanks for the change. I would suggest to print those messages to `stderr` as, especially when printing json, this might break existing scripts trying to parse the output. What do you think @smutel ?

smutel:
Sure. I updated my PR.

randombenj (approved):
LGTM

PR #1065, opened 2022-04-22
Use proper version comparisions for querys

russelltg:
Fixes package queries using string comparion instead of version comparaision

## Requirements

## Description of the Change

String comparision is not proper for version comparision!

## Checklist

- [ ] functional test added/updated (if change is functional) -- if this is needed, it would be great to get some instructions
- [x] man page updated (if applicable) N/A I beleive, bugfix
- [x] bash completion updated (if applicable) N/A
- [x] documentation updated N/A

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -46,3 +46,4 @@ List of contributors, in chronological order:
 * Benj Fassbind (https://github.com/randombenj)
 * Markus Muellner (https://github.com/mmianl)
 * Chuan Liu (https://github.com/chuan)
+* Samuel Mutel (https://github.com/smutel)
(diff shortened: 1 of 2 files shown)


randombenj:
Thanks for the changes! I would agree that string version comparison is not the right way to go. However, the code (https://github.com/aptly-dev/aptly/commit/ae3c98c210320481ad8943206d9cab6b82773add) has been in aptly for 8 years like this! I am a bit hesitant to merge this as it might break some random things ... Did you encounter any specific issues with the string version comparison?

russelltg:
Yes, I was trying to write a script to remove all versions less than the most recent version minus some number, and the subtracted number had less digits so it ended up removing many versions higher than expected (for example, with that string compassion, versions 2 > 10. That test I included fails before my changes

Issue #1060, opened 2022-04-12
`aptly publish update` fails for azure


floriankramer:
## Detailed Description
Running `aptly publish update focal azure:<name>:` fails with 
```
ERROR: unable to update: error deleting pool/main/a from Azure: https://<>.blob.core.windows.net/$web/apt/debian: -> github.com/Azure/azure-storage-blob-go/azblob.newStorageError, /go/pkg/mod/github.com/!azure/azure-storage-blob-go@v0.13.0/azblob/zc_storage_error.go:42
===== RESPONSE ERROR (ServiceCode=DirectoryIsNotEmpty) =====
Description=This operation is not permitted on a non-empty directory.
RequestId:cb72bd80-e01e-00ab-3a50-4e9ea7000000
Time:2022-04-12T09:37:31.5467576Z, Details: 
   Code: DirectoryIsNotEmpty
   DELETE https://<>.blob.core.windows.net/$web/apt/debian/pool/main/a?timeout=61
   Authorization: REDACTED
   User-Agent: [Azure-Storage/0.13 (go1.18; linux)]
   X-Ms-Client-Request-Id: [0ddb6995-a305-4d36-4c3f-1821097aaccd]
   X-Ms-Date: [Tue, 12 Apr 2022 09:37:31 GMT]
   X-Ms-Version: [2019-12-12]
   --------------------------------------------------------------------------------
   RESPONSE Status: 409 This operation is not permitted on a non-empty directory.
   Content-Length: [245]
   Content-Type: [application/xml]
   Date: [Tue, 12 Apr 2022 09:37:31 GMT]
   Server: [Windows-Azure-Blob/1.0 Microsoft-HTTPAPI/2.0]
   X-Ms-Client-Request-Id: [0ddb6995-a305-4d36-4c3f-1821097aaccd]
   X-Ms-Error-Code: [DirectoryIsNotEmpty]
   X-Ms-Request-Id: [cb72bd80-e01e-00ab-3a50-4e9ea7000000]
   X-Ms-Version: [2019-12-12]


: -> github.com/Azure/azure-storage-blob-go/azblob.newStorageError, /go/pkg/mod/github.com/!azure/azure-storage-blob-go@v0.13.0/azblob/zc_storage_error.go:42
===== RESPONSE ERROR (ServiceCode=DirectoryIsNotEmpty) =====
Description=This operation is not permitted on a non-empty directory.
[...]

PR #1069, opened 2022-04-29
Enable Azure publish unit tests in Github actions

chuan:
## Description of the Change

Enable Azure publishing unit tests in Github actions using Azure storage emulator (https://github.com/Azure/Azurite). 

## Why is this change important?

Ensure Azure publish APIs are tested in CI jobs.

Including Azurite in CI also enables adding system tests for Azure publishing functionalities in the future. Since there are already some issues (e.g., https://github.com/aptly-dev/aptly/issues/1060) reported on the Azure publish command, I plan to add system tests to cover common usages.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,6 +76,10 @@ jobs:
           pip install -U pip setuptools
           pip install -r system/requirements.txt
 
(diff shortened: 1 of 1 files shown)


randombenj (approved):
Thanks!

Issue #137, opened 2014-11-04
Can't pass any flags


kayone:
When I try to pass flags to any command I get `ERROR: unable to parse command`

`aptly repo create sonarr -distribution="master"`

gives me an error but if I remove the `-distribution="master"` it works just fine. same thing when I try to edit a repo.

I'm sure it's something really stupid that I'm missing but I've been at it for a hour and can't figure it out.

PR #1074, opened 2022-05-19
Add support for storing the "local" pool on Azure

refi64:
## Requirements

## Description of the Change

This adds support for storing the "local" package pool on Azure. It includes:

- Misc. test fixes from the Python 3 migration when passing `--capture`.
- Adding some functional tests for Azure publishing based on the S3 ones, as a foundation for these changes.
- Changes to the Azure endpoint syntax to match the official Azure connection string format.
- Some refactoring of Azure publishing to move any code reused by the package pool to a shared file.
- Tangentially related change to clean up temporary files when mirroring.

I described some of these in further detail in the individual commits.

This does *not* include changing all references to the "local" package pool, because I'm honestly not sure if that's worthwhile...

## Checklist

- [ ] ~~bash completion updated (if applicable)~~ (CLI was not touched)
- [ ] ~~documentation updated~~ (I don't believe the changes here would result in changes to the website content)


chuan (commented):
Nice work! Great to see the new functionality to store the package pool in the remote storage. I will take a more detailed pass of the change later today.

context/context.go
```diff
@@ -360,7 +360,26 @@ func (context *AptlyContext) PackagePool() aptly.PackagePool {
...
+			}
+		} else {
+			poolRoot := context.config().PackagePoolStorage.Path
+			if poolRoot == "" {
+				poolRoot = filepath.Join(context.config().RootDir, "pool")
+			}
+
+			context.packagePool = files.NewPackagePool(poolRoot, !context.config().SkipLegacyPool)
+		}
 	}
```
chuan, line 383: Can we make the package pool storage configs similar to publish configs and let user make the choice instead of preferential based selection here? For example, the users can have `FileSystemPackagePoolRoots` and `AzurePackagePoolRoots` with settings similar to `FileSystemPublishRoots` and `AzurePublishRoots`. The `PackagePoolStorage` will select an existing config. I think this will the config less confusing to the users.

refi64: I will admit, I'm kind of mixed on the current configuration approach :sweat_smile: My rationale was two-fold:

- It shouldn't look like it's easy / workable to randomly change package pools, once it's set to filesystem or azure, it's basically *set*, unlike publishing where you can decide to just publish somewhere else
- Having parallel `AzurePackagePoolRoots` and `AzurePublishRoots`, both probably containing identical key-value pairs, feels a bit ugly

So I was split between two solutions:

- Rename the `*PublishRoots` to, say, `*Endpoints`, and have the `PackagePoolStorage` reference that
- The current solution

and I just went over the current approach due to the first point, but again I'm not really a huge fan / not tied to it. Would going with the other solution also work? (Presumably backwards-compat would be kept, of course! Having the configuration split between two likely near-identical places just feels off IMO...even though that's sort of the current solution anyway I guess?)

chuan: I understand your concern. The way I read the current solution with multiple choices is that both the local path and remote will be used as package pool storage instead of picking one over another. Probably we should either err on more than one config or actually store the packages on all configured storage pools. What do you think?

azure/package_pool.go
```diff
@@ -0,0 +1,240 @@
...
+		if err != nil {
+			return "", err
+		}
+	}
+
+	path := pool.buildPoolPath(basename, checksums)
+	blob := pool.az.blobURL(path)
+
+	var leaseID string
+	lease, err := blob.AcquireLease(context.Background(), "", leaseDuration, azblob.ModifiedAccessConditions{})
```
chuan, line 168: Is the lease strictly needed? Without periodically renewing the lease, this will set the timeout of the operation to the lease duration from my understanding.

azure/package_pool.go
```diff
@@ -0,0 +1,240 @@
...
+
+func (pool *PackagePool) Open(path string) (aptly.ReadSeekerCloser, error) {
+	blob := pool.az.blobURL(path)
+
+	temp, err := ioutil.TempFile("", "blob-download")
+	if err != nil {
+		return nil, errors.Wrap(err, "error creating temporary file for blob download")
+	}
+
+	defer os.Remove(temp.Name())
```
chuan, line 114: Won't this delete the downloaded temp file upon return of the function? I thought the caller might still want to read the content of the downloaded temp file.

context/context.go
chuan: I think either `type` or `AzureEndpoints` works. 

>  even though you can't switch package pools right now
Can I upload an existing local package pool to the Azure blob storage and set the pool config to the storage account to make the switch?

refi64: > Can I upload an existing local package pool to the Azure blob storage and set the pool config to the storage account to make the switch?

ATM the paths are different, but I can make them consistent with the normal package pool and then it should work. (Technically, the legacy path locations won't work, but I'm not sure if that's something worth adding.)

azure/package_pool.go
refi64: Hmm, it was intended to serve as an "operation lock" of sorts...but I think I was indeed over-eager with the use of leases, since many of this wouldn't have particularly bad behavior if run in parallel (e.g. the same file gets uploaded twice, but Azure won't let the uploads corrupt each other already, so it wouldn't be an issue). I'll remove the lease usage (probably some other areas in the file I can remove it from as well).

azure/package_pool.go
refi64: On Posix systems (Linux, macOS), you can still interact with a file handle after you've deleted it; the backing is still there, the on-disk "handle" for it was just removed. It's truly deleted once the file handle is closed. (This is a common trick in the temporary file functions of some languages, in particular Python's `tempfile` does this, as does Rust's `tempfile` crate.)

refi64:
Hmm, the Azure logs seem to have some 500 internal server errors from Azurite, I'll look into it locally.

context/context.go
chuan: I don't think we need to support the legacy path. But making both package pool paths consistent is definitely a huge plus, i.e. users can just upload the existing local package pool to Azure and update the config to switch to the remote package pool.

refi64:
I can't reproduce the CI failure at all locally, so I added a step to upload Azurite's logs if the tests fail, which should hopefully help clear things up a bit...

chuan (approved):

refi64:
```
Can't use 'tar -xzf' extract archive file: /home/runner/work/_actions/_temp_4c5950e6-f316-4a7c-aa09-7b03239b3a59/3b979214-6f9f-46a7-976a-ed55121cf5e1.tar.gz. return code: 2.
```

...GitHub Actions issue maybe?

chuan:
> ```
> Can't use 'tar -xzf' extract archive file: /home/runner/work/_actions/_temp_4c5950e6-f316-4a7c-aa09-7b03239b3a59/3b979214-6f9f-46a7-976a-ed55121cf5e1.tar.gz. return code: 2.
> ```
> 
> ...GitHub Actions issue maybe?

Where did you see this error?

Since in the [CI #137 run](https://github.com/aptly-dev/aptly/actions/runs/2352295323) only the long test for v1.17 failed, I suspect it is an intermediate issue with Azurite or the runtime environment.

refi64:
@chuan apologies for the confusion, that was from an earlier run, but I think @randombenj restarted it since then. There was a bug in the way config files were written that I missed, just pushed up a fix.

chuan:
@refi64 I figured out that the Azurite 500 errors are due to its use of `/tmp`. Your change should pass the CI if you pull the change here: https://github.com/aptly-dev/aptly/pull/1078.

refi64:
@chuan thanks! I just pushed a new version including that commi

.github/workflows/ci.yml
```diff
@@ -87,12 +89,19 @@ jobs:
...
           AZURE_STORAGE_ACCESS_KEY: "Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw=="
         run: |
           make
 
+      - name: Upload Azurite logs
+        if: failure()
+        uses: actions/upload-artifact@v3
+        with:
+          name: azurite-logs
+          path: /tmp/azurite/debug.log
```
chuan, line 103: This should be `${{ runner.temp }}/azurite/debug.log` given the directory setting above.

aptly/interfaces.go
```diff
@@ -35,8 +35,8 @@ type PackagePool interface {
 	Import(srcPath, basename string, checksums *utils.ChecksumInfo, move bool, storage ChecksumStorage) (path string, err error)
 	// LegacyPath returns legacy (pre 1.1) path to package file (relative to root)
 	LegacyPath(filename string, checksums *utils.ChecksumInfo) (string, error)
-	// Stat returns Unix stat(2) info
-	Stat(path string) (os.FileInfo, error)
+	// Size returns the size of the given file.
```
adalessandro, line 38:
```suggestion
	// Size returns the size in bytes of the given file.
```

system/t02_config/ConfigShowTest_gold
```diff
@@ -15,6 +15,7 @@
     "gpgDisableVerify": false,
     "gpgProvider": "gpg",
     "downloadSourcePackages": false,
+    "packagePoolStorage": {},
```
chuan, line 18: I think you need to add `"type": "local"` here or update the code to default to empty and add code to deal with empty config. In the form of current code, this will fail the test due to mismatching between `config.go` and content here. Same for `CreateConfigTest_gold` below.

refi64: Ack, that's what I get for not re-running the full test suite when I changed the config. As-is it already handles this being blank, so I just updated it to not bother adding anything to it if everything is still in the default values.

chuan (approved):

chuan:
The latest change looks good to me and glad to see all CI jobs passed!

refi64:
Just rebased on top of the merged master incl. the Azurite fixes PR.

randombenj:
As this is quite a big change and I am not too familiar with the source yet, this will probably take some time to review ...

randombenj:
I am not sure if I understand all the changes yet ...
However, I am wondering a bit why the changes are necessary in the first place?
Couldn't you just mount the azure blob storage to the local disk and then run aptly as is to this directory?
This would be much simpler and aptly wouldn't have to implement all the read write logic?

refi64:
As an aside, in addition to the other reasons already detailed above,
using FUSE filesystems for stuff like this tends to introduce a decent
number of new and exciting bugs, simply because the ways that the
filesystems deviate from standard POSIX behavior usually aren't
accounted for in various tools.

Quoting Benj Fassbind (2022-06-23 04:17:49)

randombenj (approved):
Thanks for this high quality patch! I think it also allows for future implementations in different storage backends.

One thing I was wondering however, what is your main use case for implementing this changes?

azure/package_pool_test.go
```diff
@@ -0,0 +1,262 @@
...
+	c.Assert(err, IsNil)
+	cnt := s.pool.az.container
+	_, err = cnt.Create(context.Background(), azblob.Metadata{}, azblob.PublicAccessContainer)
+	c.Assert(err, IsNil)
+
+	s.prefixedPool, err = NewPackagePool(s.accountName, s.accountKey, container, prefix, s.endpoint)
+	c.Assert(err, IsNil)
+
+	s.checksum = utils.ChecksumInfo{
+		MD5: "0035d7822b2f8f0ec4013f270fd650c2",
```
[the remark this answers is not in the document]
refi64: Looks like this was a weird leftover from me copy-pasting the initial tests from the ones for the local package pool :sweat_smile: It's still not really clear to me why this was here (one test seems to use it, but...it could've just been in that test), but for here, I removed this, *and* also took the opportunity to clean the handling of checksums throughout the tests.

azure/package_pool_test.go
```diff
@@ -0,0 +1,262 @@
...
+	delete(s.cs.(*files.MockChecksumStorage).Store, path)
+	s.checksum.SHA512 = ""
+	ppath, exists, err = s.pool.Verify("", filepath.Base(s.debFile), &s.checksum, s.cs)
+	c.Check(ppath, Equals, path)
+	c.Check(err, IsNil)
+	c.Check(exists, Equals, true)
+	// checksum is filled back based on re-calculation
+	c.Check(s.checksum.SHA512, Equals, "d7302241373da972aa9b9e71d2fd769b31a38f71182aa71bc0d69d090d452c69bb74b8612c002ccf8a89c279ced84ac27177c8b92d20f00023b3d268e6cec69c")
+
+	// check existence, with wrong size
```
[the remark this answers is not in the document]
refi64: forgot to add it back :sweat_smile:

azure/package_pool_test.go
```diff
@@ -0,0 +1,262 @@
...
+	c.Check(err, IsNil)
+	c.Check(exists, Equals, false)
+}
+
+func (s *PackagePoolSuite) TestImportNotExist(c *C) {
+	_, err := s.pool.Import("no-such-file", "a.deb", &s.checksum, false, s.cs)
+	c.Check(err, ErrorMatches, ".*no such file or directory")
+}
+
+// func (s *PackagePoolSuite) TestImportOverwrite(c *C) {
```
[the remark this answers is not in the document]
refi64: afaik it's another weird holdover from the files package pool tests? The test itself didn't really make sense to me: it seemed to be checking what would happen if a file was already at the same path with different checksums, but we check that elsewhere in the Azure pool anyway, so it just doesn't work here.

Leaving it commented-out was a mistake though, fixed.

system/azure_lib.py
```diff
@@ -0,0 +1,111 @@
...
+        )
+    else:
+        print('Azure tests disabled: Azure creds not found in the environment')
+        blob_client = None
+except ImportError as e:
+    print("Azure tests disabled: can't import azure.storage.blob", e)
+    blob_client = None
+
+
+class AzureTest(BaseTest):
```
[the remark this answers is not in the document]
refi64: Just added checks for that!

neolynx:
could you check if https://github.com/aptly-dev/aptly/pull/1283 works for you ?

especially since a prefix was introduced in azure/public.go ...

neolynx (approved):
tested also manually

PR #1079, opened 2022-06-12
api: allow parameters with urlencoded names

sjoerdsimons:
Aptly allows create e.g. repos with a / to use those with the REST api
the router needs to allow urlencoded parameters in various places to
represent this. A specific example of this is the /api/repos/:name/packages path


randombenj:
So this allows to have urlencoded `/` in the `:name` variable in for example `/api/repos/:name/packages`?

randombenj (approved):

PR #1081, opened 2022-06-18
Allow disabling bzip2 compression for index files

sjoerdsimons:
Using bzip2 generates smaller index files (roughly 20% smaller Packages
files) but it comes with a big performance penalty.  When publishing a
debian mirror snapshot (amd64, arm64, armhf, source) without contents
skipping bzip speeds things up around 1.8 times.

```
$ hyperfine -w 1 -L skip-bz2 true,false  -m 3 -p "aptly -config aptly.conf publish drop bullseye || true" "aptly -config aptly.conf  publish snapshot  --skip-bz2={skip-bz2} --skip-contents --skip-signing bullseye"
Benchmark 1: aptly -config aptly.conf  publish snapshot  --skip-bz2=true --skip-contents --skip-signing bullseye
  Time (mean ± σ):     35.567 s ±  0.307 s    [User: 39.366 s, System: 10.075 s]
  Range (min … max):   35.311 s … 35.907 s    3 runs

Benchmark 2: aptly -config aptly.conf  publish snapshot  --skip-bz2=false --skip-contents --skip-signing bullseye
  Time (mean ± σ):     64.740 s ±  0.135 s    [User: 68.565 s, System: 10.129 s]
  Range (min … max):   64.596 s … 64.862 s    3 runs

Summary
  'aptly -config aptly.conf  publish snapshot  --skip-bz2=true --skip-contents --skip-signing bullseye' ran
    1.82 ± 0.02 times faster than 'aptly -config aptly.conf  publish snapshot  --skip-bz2=false --skip-contents --skip-signing bullseye'
```

Allow skipping bz2 creation for setups where faster publishing is more
important then Package file size.


sjoerdsimons:
> Thanks for this change, it looks good in itself! I am wondering however, if bz2 compression is 'optional' as this change suggests, why is it used by some apt repos? Will this have any implications on how apt queries those repos?

Apt first downloads the Release or InRelease files; Which will list all the other index files; From there it will pick the most suitable one (e.g. Package, Packages.gz etc) for a given component/architecture.

For reference Debian itself doesn't use .bz2 compressed Packages files anymore. Just uncompressed, gzip and xz (see e.g. http://ftp.debian.org/debian/dists/bullseye/Release)

The background of my various patches is that we're for Apertis.org we're moving from reprepro to aplty.  Reprepro, or at least the version we used, also doesn't support bzip compressed packages files and has been working just fine :). 

Hence not wanting to spent lots of CPU cycles on bzip2 now.. For the future adding xz compression may be more interesting as that provides a better compression ration. But really size of Packages files is not something that is a high priority

randombenj (approved):

itayporezky:
This was merged and running on aptly 1.5.0, but this isn't documented on the website
`-skip-bz2` flag should be documented on https://www.aptly.info/doc/aptly/publish/repo/ and other publish commands

Issue #1071, opened 2022-05-16
Publish could be faster by parallelizing uncompress threads


Hybrid512:
Aptly is great but publishing snapshots can take ages with big mirrors when generating metadata.
Adding more threading capacity (by adding more CPU cores for example) doesn't help by much since uncompress is done *apparently* in a single thread and this is especially slower with newer repositories based on zstd compression which takes *a lot* longer to uncompress than tar.gz.

![image](https://user-images.githubusercontent.com/18528623/168582444-6c25e451-b636-4f72-ad67-60b8b542d31e.png)

Switching to more uncompressing threads (like N-1 threads for N vCPUs so that we still keep 1 vCPU for the system responsiveness) would probably greatly boost publish speed.

## Your Environment
Aptly master branch : aptly version: '1.4.0+179+g0a179886'
OS : Ubuntu 20.04 amd64
Storage: Filesystem (ext4)
HW specs : KVM VM, host-passthrough CPU model (same as host, Intel Xeon Gold 6242 CPU @ 2.80GHz) 8 vCPUs, 4Gb RAM, 500GB disk
Mirrors : ubuntu 20.04 and 22.04 (main, updates, security with main,restricted,universe,multiverse components)

PR #1082, opened 2022-06-18
Use parallel gzip instead of gzip for compression

sjoerdsimons:
golangs compress/gzip isn't a parallel implementation, so it's quite a
bit slower on most modern servers then pgzip. The below benchmark
run shows that publishing a debian bullseye mirror snapshot (amd64, arm64,
armhf, source) shows a gain of about 35% in publishing time (when skipping
bz2 using MR #1081)

```
 hyperfine -w 1 -m 3 -L aptly aptly-nobz2,aptly-nobz2-pgzip -p "{aptly} -config aptly.conf publish drop bullseye || true" "{aptly} -config aptly.conf  publish snapshot --skip-bz2=true --skip-contents --skip-signing bullseye"
Benchmark 1: aptly-nobz2 -config aptly.conf  publish snapshot --skip-bz2=true --skip-contents --skip-signing bullseye
  Time (mean ± σ):     35.548 s ±  0.378 s    [User: 39.465 s, System: 10.046 s]
  Range (min … max):   35.149 s … 35.902 s    3 runs

Benchmark 2: aptly-nobz2-pgzip -config aptly.conf  publish snapshot --skip-bz2=true --skip-contents --skip-signing bullseye
  Time (mean ± σ):     26.592 s ±  0.069 s    [User: 42.207 s, System: 9.676 s]
  Range (min … max):   26.521 s … 26.660 s    3 runs

Summary
  'aptly-nobz2-pgzip -config aptly.conf  publish snapshot --skip-bz2=true --skip-contents --skip-signing bullseye' ran
    1.34 ± 0.01 times faster than 'aptly-nobz2 -config aptly.conf  publish snapshot --skip-bz2=true --skip-contents --skip-signing bullseye'
```


randombenj:
Is this a fix for https://github.com/aptly-dev/aptly/issues/1071 ?

sjoerdsimons:
> Is this a fix for #1071 ?

I think #1071 was specifically about speeding up .deb *decompression* by using parrallelism when creating content files.  Where as this change is parallelizing compression of the index files (e.g. Packages.gz). So they seem orthogonal to me

sjoerdsimons:
Fwiw to fix the merge conflict pushed in an extra commit to run `go mod tidy` first to avoid having odd changes in the actual functional commit

randombenj (approved):

neolynx:
I think pgzip is not used everywhere, need to check also about xz...

Issue #1067, opened 2022-04-27
Aptly does not support publishing objects without ACL rules


mdebord-dlr:
## Description

Amazon now suggests to DISABLE ACL's for object ownership.
It's even marked as "**recommended**" in the AWS S3 console. 

Aptly does not support publishing objects without ACL rules. When I attempt to publish to a bucket with this recommended setting, it throws `AccessControlListNotSupported: The bucket does not allow ACLs status code: 400`

See: https://aws.amazon.com/about-aws/whats-new/2021/11/amazon-s3-object-ownership-simplify-access-management-data-s3/ for more info. 

## Context
I originally found this when my account resources couldn't pull objects from S3 that were published by Aptly in a different account. So I investigated permissions and found that the account that published the objects to S3 owned them, and regardless of the bucket policy permissions, resources within the bucket owner account could not retrieve the files. 

My use-case has a single Aptly instance that manages snapshots for all environments, it then pushes objects into S3 buckets in **various accounts**. It publishes from this single Aptly instance to ensure snapshots are identical across all environments/accounts. 

The ability to publish from a single mirror repo to many S3 publish endpoints across accounts is a useful functionality and mission-critical for my use-case. 

## Possible Implementation
- Do not publish any `acl` information when publishing objects to S3 if the `acl` parameter is missing from config. Alternatively, to support backwards compatibility, leave `private` acl as default and support a null/None keyword to tell Aptly to ignore setting `acl` when publishing objects.

wadey:
I ran into the same issue, so I opened a PR to fix it: #1087

PR #1087, opened 2022-06-22
S3: support disabling ACL with `none` value

wadey:
This change lets you disable ACL when using S3 by using a configuration
value of `none`. This way we maintain backward compatibility with the
default setting being `private`.

Fixes: #1067

## Checklist

- [ ] documentation updated

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -48,3 +48,4 @@ List of contributors, in chronological order:
 * Chuan Liu (https://github.com/chuan)
 * Samuel Mutel (https://github.com/smutel)
 * Russell Greene (https://github.com/russelltg)
+* Wade Simmons (https://github.com/wadey)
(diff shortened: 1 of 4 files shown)


randombenj (approved):

wadey:
@randombenj ah I found it, PR opened:

- https://github.com/aptly-dev/www.aptly.info/pull/97

AirOne70:
This fix is not working on my side. I had to use " " instead of "none" to make it working.
From what I can see with Debug enabled "X-Amz-Acl: none" is set but S3 answer is "HTTP/1.1 400 Bad Request".

wadey:
@AirOne70 are you sure you are running aptly version 1.5 or newer? Can you check your version?

AirOne70:
@wadey Thanks for your answer. You're right, I updated Aptly only few days before the 1.5 release. That explains it. Sorry about that!

PR #1088, opened 2022-06-22
feat: repo and snapshots packages filter api add 'maximumVersion' que…

hudeng-go:
repo and snapshots packages filter api add 'maximumVersion' query parameter support


randombenj:
Could you please add a test for the suggested changes?

goldendeng:
> Could you please add a test for the suggested changes?

Added an API test.

neolynx:
continued in https://github.com/aptly-dev/aptly/pull/1301

Issue #115, opened 2014-10-02
Support for '/' in distribution names


smira:
Requires special component naming, etc. To be investigated.

Requested by @erickeller in #110.

sandrotosi:
yeah exactly, / is replaced with - so we cannot mimicwhat Debian is doing for the security updates, where the distribution is set to 'wheezy/updates' but with aptly we ended up with "wheezy-security", but what we really want is 'wheezy/updates'

please allow / in distribution name

wwarmus:
Hey, I think same issues concerns the prefix values. 
While trying to publish via api I'm hitting 404, but can publish via cli, though...
Is it possible to add support for /api for the :prefix names containing "/" too?

rottenbytes:
:+1: on this one. allowing "/" in names via api would be great.

richard-scott:
Looking forward to testing this one as it is the last thing I need to have working for aptly to be the perfect tool for my needs :-)

mauricemeyer:
This relates to #561.

ksalman:
This would be nice to have.

richard-scott:
Wow, I'd forgotten about this tool.... 
Will this feature get into v1.3.0?

Would be great if it did.

YotamAlon:
Could also use this feature

adalessandro:
Following up on this thread, what's the reason to forbid slash `/` characters in the publish distribution name? I.e. I'm trying to understand this line of code: https://github.com/aptly-dev/aptly/blob/v1.4.0/deb/publish.go#L265

PR #1089, opened 2022-06-24
RFC: Allow '/' in distribution name

adalessandro:
This is an RFC PR to follow up on the ongoing opened issue https://github.com/aptly-dev/aptly/issues/115
Possible duplicated or already a declined feature. But it's not really clear why this is considered a bug in aptly and what would be the consequences of allowing slash `/` characters in the distribution name.

This reverts commit 1daa076d65e642f2f365e3cd580a78548a9b65a6.

Fixes https://github.com/aptly-dev/aptly/issues/115

## Requirements

## Checklist

- [ ] documentation updated

### DIFF
--- a/deb/publish.go
+++ b/deb/publish.go
@@ -197,9 +197,6 @@ func NewPublishedRepo(storage, prefix, distribution string, architectures []stri
 		if distribution == "" || component == "" {
 			rootDistributions, rootComponents := walkUpTree(source, collectionFactory)
 			if distribution == "" {
-				for i := range rootDistributions {
-					rootDistributions[i] = strings.Replace(rootDistributions[i], "/", "-", -1)
-				}
 				discoveredDistributions = append(discoveredDistributions, rootDistributions...)
 			}
 			if component == "" {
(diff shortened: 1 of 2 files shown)


sjoerdsimons:
For reference we've been running with this revert for Apertis for quite a while now and things seem entirely happy ;) But as @adalessandro mentioned we may well be missing the background reason of why this was disallowed  in the first place?

sjoerdsimons:
fwiw the one thing we noticed recently  that when publishing a distribution with a `/` "under" another distribution then removing the top-level one will remove everything underneath :) So that was likely based on the assumption that distribution names under a prefix don't have a `/` (I'm not sure if aptly also protects against nested prefixes)

As a more practical example of what i meant above, for Apertis, we publish both  the current  repositories and snapshots with a setup as follows:
* current: `/apertis/dists/v2024pre`
* snapshot: `/apertis/dists/v2024pre/snapshots/<timecode>`

When dropping current it will actually wipe out everything under `/apertis/dists/v2024pre` which ofcourse then includes all published snapshots. This was only uncovered recently due to some transition/testing as we'd normally never want to delete the current published repositories and leave the snapshots :)

So for this RFC that's one thing to fix for sure; Based on our experiences thusfar that's likely the only issue with this RFC

neolynx:
are you creating local repos with a / in the distributions, or do you use mirrors with such ?

neolynx:
superseded by https://github.com/aptly-dev/aptly/pull/1269

PR #1093, opened 2022-07-18
fix:failed to edit repo

cccccarl:
return when  the repo cant be found and err !=nil

Fixes #

## Requirements

## Description of the Change

The repo cant be edited 

## Checklist

- [ ] documentation updated


randombenj:
I don't quite get what's the problem here, I would expect that when the repo already exists with the same name that it is not renamed and this is exactly how it is implementad at the moment.
This is also the same behaviour as in the cli: https://github.com/aptly-dev/aptly/blob/master/cmd/repo_rename.go#L30

PR #1100, opened 2022-08-14
api: Allow querying the packages endpoint

sjoerdsimons:
The ".../packages" endpoints for mirror, local repos and snapshots all
share the same syntax for querying. However the "/api/packages" endpoint
doesn't match this. Adjust that to allow for a bit more consistency and
allow querying the full package database.

The current endpoint functionality "/packages/:name" is kept intact and
can be used the same as now

### DIFF
--- a/api/packages.go
+++ b/api/packages.go
@@ -15,3 +15,10 @@ func apiPackagesShow(c *gin.Context) {
 
 	c.JSON(200, p)
 }
+
+// GET /api/packages
+func apiPackages(c *gin.Context) {
+	collectionFactory := context.NewCollectionFactory()
+	collection := collectionFactory.PackageCollection()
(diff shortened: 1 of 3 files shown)


randombenj:
@sjoerdsimons The CI should be fixed again, can you rebase your branch?

sjoerdsimons:
Bit surprised of the code coverage report here as the code is tested in the python based systemd testing.  I guess the coverage only detects the go based tests or ?

randombenj (approved):
LGTM! Thanks for the awesome improvement!

randombenj:
@sjoerdsimons regarding test coverage: python test should be recorded in the coverage ... but there seems some issue with that ... I will investigate in some later change

Issue #1017, opened 2022-02-04
Serve api and published packages simultaneously


aol-nnov:
Not sure about the reason making it two separate commands, but the following patch allows to perform both in one call.

```patch
diff --git a/api/router.go b/api/router.go
index fdfbb94..e164d97 100644
--- a/api/router.go
+++ b/api/router.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+ "github.com/aptly-dev/aptly/aptly"
        "net/http"
 
        ctx "github.com/aptly-dev/aptly/context"
@@ -15,6 +16,8 @@ func Router(c *ctx.AptlyContext) http.Handler {
 
        router := gin.Default()
        router.Use(gin.ErrorLogger())
+ publicPath := context.GetPublishedStorage("").(aptly.FileSystemPublishedStorage).PublicPath()
+ router.Static("/repos/", publicPath)
 
        if context.Flags().Lookup("no-lock").Value.Get().(bool) {
                // We use a goroutine to count the number of
```

Changes are pretty minimalistic and I'd be glad to see them included.
If there are no objections, I'm ready to make a PR. (disclamer: I'm not a gopher ;) )

Issue #975, opened 2021-08-31
aptly CLI should be able to work while aptly web is running


lbolla:
At the moment, one has to stop the aptly web server in order to be able to use the `aptly` command from the terminal. This is because leveldb prevents multiple processes from accessing the same database.

We should lift this limitation. One obvious way would be for the CLI to make use of the API (although this would require the API to be running when using the CLI, which is not always wanted/possible).

lbolla:
I found few API-based aptly clients already, e.g. https://github.com/gopythongo/aptly-api-client

umlaeute:
afaiu, there are many many commands of the `aptly` command unavailable via the API.

e.g. with `snapshot` the API provides (`list`, `show`, `create` (from local repo resp pkg ref), `update`, `delete`) whereas `aptly snapshot` additionally provides `verify`, `pull`, `filter`, `diff`, `merge`, `search`.
are all these just syntactic sugar?

also, there is no way to do an `aptly db cleanup` via the API.

it would be great if the API would have a feature parity with the cmdline.

(i'm referring mostly to `aptly` as documented on https://www.aptly.info/doc; probably the development branches already have all this)

umlaeute:
actually i found https://www.aptly.info/doc/api/ soon afterwards (obviously i only read the specific API endpoint documentation, but not the general one), which explains how to emulate many of the missing commands.
i can now do most of my jobs in bash with just `curl` and `jq`, so you could mainly consider my comment obsolete.

having a way to run `db cleanup` via the API would still be very nice.

flotho:
Hi,
I'm unable to find the proper resource (I still looking for... ;-) ) but if you run an nginx or an apache websever pointing at the root of your aptly path, you'll be able to replace aptly serve.
IMHO it would be more efficient to try out this one so you'll be able to keep benefits of both functions without overloading dev team that re-handle this awesome project  .
Did you already had a try on this kind of conf ?

regards

flotho:
additionnaly could we say that  #1017 could be an answer to your needs ?
@randombenj , here is probably some duplicates to clean up.
also maybe I'm right and not able to point the proper documentation item for :
>  but if you run an nginx or an apache websever pointing at the root of your aptly path, you'll be able to replace aptly serve.

PR #1116, opened 2022-09-24
Serve api and published repos simultaneously

aol-nnov:
It all has started with a small patch in #1017 which did not gain much interest, unfortunately.
Later #975 was linked and yet, no PR.

I've got a bit tired rebasing this small fix every time I build a docker container and I have no time to maintain a fork, so here we are. :)

These changes were battle-tested for more than 6 months now. Please consider including them into the mainline.

Full disclosure: I'm not a golang developer ;)


aol-nnov:
Hello, anyone? :)

@randombenj any chance you could ping someone who will look into this PR, please? )

api/router.go
```diff
@@ -21,6 +23,7 @@ func Router(c *ctx.AptlyContext) http.Handler {
 	context = c
 
 	router := gin.Default()
+	router.RedirectTrailingSlash = false
```
randombenj, line 26: Couldn't this be potentially breaking for some users?

randombenj:
The changes look good to me, would you mind adding a small system test so we can verify the functionality and that this doesn't break in future chagnes?

api/router.go
aol-nnov: Might be, but, unfortunately, without it it does not work for files.. All requests get a trailing slash and `FileFromFS` fails..

aol-nnov:
> would you mind adding a small system test 

Unfortunately, I'm not that good in golang in general and in its testing infrastructure.. I'd be glad if someone could lend me a hand with that...

api/router.go
randombenj: Maybe this could be configured just for the `/repos` endpoint?
Also it would be nice if this could be enabled/disabled via configuration as some people may not want the files to be served when the api is running?

aol-nnov: I did my best grepping gin documentation, but it seems to be a global parameter.

randombenj: Hard to say what kind of implications this will have, but the tests seem to pass so at least it doesnt seem to break ...

randombenj:
@aol-nnov The system tests are in python: https://github.com/aptly-dev/aptly/blob/master/system/t12_api/publish.py
Maybe you are more familiar with that? Otherwise I can have a look at this.

aol-nnov:
> Otherwise I can have a look at this.

I'd be glad if you will, @randombenj 😊

randombenj:
@aol-nnov I will see when I have time for this, but I will try to take this up :)

mmianl:
@randombenj, @aol-nnov 

Is PR #1166 what you had in mind?

aol-nnov:
> what you had in mind

moreover, @randombenj , it's my patchset verbatim! (plus system tests) :-P
But whatever. I'd be glad if it will land in the mainline either way.

Still, I have a question about `RedirectTrailingSlash` param. If I recall it right, it does not work for me without it.

mmianl:
Regarding `RedirectTrailingSlash`: I don't see any issues with it not being set to `false`. In fact, setting it to `false` breaks other tests...

aol-nnov:
@mmianl if my memory serves me right, requests to files end up with trailing slash in the end of the path (and they should not) if `RedirectTrailingSlash` is not set to false.

mmianl:
@aol-nnov can you try again from my branch and see if it works?

aol-nnov:
@mmianl well, it appeared, there is no redirect when file is requested:

```
aol@Andreys-MacBook-Pro:~/.aptly$ curl -v http://localhost:8080/repos/bootstrap/pool/main/t/time/time_1.9-0.1_amd64.deb
*   Trying 127.0.0.1:8080...
* Connected to localhost (127.0.0.1) port 8080 (#0)
> GET /repos/bootstrap/pool/main/t/time/time_1.9-0.1_amd64.deb HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.86.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Accept-Ranges: bytes
< Content-Length: 50768
< Content-Type: application/vnd.debian.binary-package
< Last-Modified: Fri, 15 Jan 2021 11:58:44 GMT
< Date: Mon, 20 Mar 2023 12:15:01 GMT
< 
Warning: Binary output can mess up your terminal. Use "--output -" to tell 
Warning: curl to output it to your terminal anyway, or consider "--output 
Warning: <FILE>" to save to a file.
* Failure writing output to destination
* Closing connection 0
```

It's been a while since I made this patchset and unfortunately, I can't tell now why there was a `RedirectTrailingSlash` set to false.

LGTM then :)

PR #1130, opened 2022-11-30
Add api publish mutex

### LINKED ISSUE #1125: API: failures and race conditions with concurrent publish update operations
## Detailed Description

Using Aptly 1.5.0.
Running 2 API calls `PUT {api_url}/api/publish/{path}/{distribution}` on the same path at the same time, will lead to various failures and potentially an unpredictable end result. It seems as if there is no locking at all on this operation?

This looks like a regression introduced in the latest Aptly release. I cannot remember having encountered this while we were using version 1.4.0.

## Context

We run builds on a Jenkins server that publish Debian packages to repositories. Often there are multiple builds that want to publish their different products to the same repository. Since there are multiple workers, it is not unusual that 2 builds try to upload their products and do a publish update at the same time. Often the build will fail in this case, which is already annoying enough on its own. But it gets worse: sometimes the API calls don't even seem to return an error and everything seems to be fine, but one of the packages is not in the repository. This will then lead to weird errors later on and a lot of wasted time figuring out what exactly happened.

## Your Environment

Attached is a Python3 script with some Debian packages randomly taken from Ubuntu jammy that reproduced this problem on the first try for me.
[aptly-publish-bug.tar.gz](https://github.com/aptly-dev/aptly/files/9869531/aptly-publish-bug.tar.gz)

1. Create and publish a repository `test-aptly-race-bug`.
   I have also added a random package to it, just to ensure the following steps will start from a similar state.
   `aptly repo create test-aptly-race-bug`
   `aptly repo add test-aptly-race-bug vim_8.2.3995-1ubuntu2_amd64.deb`
   `aptly publish repo -distribution jammy test-aptly-race-bug test-aptly-race-bug`
2. Open a terminal, unpack the tarball, and `cd` to the `gnome-shell` directory.
3. In this terminal, have this command line ready but don't execute it yet:
   `../publish_pkgs.py -a 'http://your-aptly-server:8080' test-aptly-race-bug jammy gnome-shell_42.4-0ubuntu0.22.04.1.dsc gnome-shell_42.4-0ubuntu0.22.04.1_source.changes`
   (Adjust the `-a` argument to point to your aptly test server, or edit `DEFAULT_APTLY_URI` in the script itself and omit this parameter.)
4. Open another terminal and `cd` to the `tzdata` directory.
5. In this terminal, have this command line ready:
   `../publish_pkgs.py -a 'http://your-aptly-server:8080' test-aptly-race-bug jammy tzdata_2022e-0ubuntu0.22.04.0.dsc tzdata_2022e-0ubuntu0.22.04.0_source.changes`
   (Again, update or omit `-a` argument.)
6. Hit **enter** in the `gnome-shell` terminal, and then as quickly as possible in the `tzdata` terminal as well, such that the 2 instances of the script start almost at the same time.
7. Kaboom 💥.

Journalctl on the aptly server during the failure:
```
Oct 26 13:37:30 mocha aptly[31969]: Signing file 'Release' with gpg, please enter your passphrase when prompted:
Oct 26 13:37:30 mocha aptly[31969]: Signing file 'Release' with gpg, please enter your passphrase when prompted:
Oct 26 13:37:30 mocha aptly[31969]: Clearsigning file 'Release' with gpg, please enter your passphrase when prompted:
Oct 26 13:37:30 mocha aptly[31969]: Clearsigning file 'Release' with gpg, please enter your passphrase when prompted:
Oct 26 13:37:30 mocha aptly[31969]: [GIN] 2022/10/26 - 13:37:30 | 500 |  310.528273ms |   192.168.12.33 | PUT      "/api/publish/test-aptly-race-bug/jammy"
Oct 26 13:37:30 mocha aptly[31969]: Error #01: unable to update: unable to rename: rename /mnt/dev-artifact/backup/mocha/public/test-aptly-race-bug/dists/jammy/main/Contents-amd64.tmp.gz /mnt/dev-artifact/backup/mocha/public/test-aptly-race-bug/dists/jammy/main/Contents-amd64.gz: no such file or directory
Oct 26 13:37:30 mocha aptly[31969]: [GIN] 2022/10/26 - 13:37:30 | 500 |  325.660056ms |   192.168.12.33 | PUT      "/api/publish/test-aptly-race-bug/jammy"
Oct 26 13:37:30 mocha aptly[31969]: Error #01: unable to update: unable to rename: rename /mnt/dev-artifact/backup/mocha/public/test-aptly-race-bug/dists/jammy/main/binary-amd64/Packages.tmp.bz2 /mnt/dev-artifact/backup/mocha/public/test-aptly-race-bug/dists/jammy/main/binary-amd64/Packages.bz2: no such file or directory
Oct 26 13:37:30 mocha aptly[31969]: [GIN] 2022/10/26 - 13:37:30 | 200 |   69.542796ms |   192.168.12.33 | DELETE   "/api/files/upload_bxvkuihc"
Oct 26 13:37:30 mocha aptly[31969]: [GIN] 2022/10/26 - 13:37:30 | 200 |   73.519901ms |   192.168.12.33 | DELETE   "/api/files/upload_cmttlfyd"
```
It is obvious that the 2 instances of the API call are just running concurrently and get in each other's way because one instance has already moved/deleted files that the other is also using. The specific failure may vary depending on the timing of starting both instances of the script. Sometimes the API call does not break entirely, but the final result lacks one of the packages because one instance overwrote the modified Packages file from the other.

randombenj:
At the moment it is posible to have multiple publish updates running at the same time. This is a problem as both processes try to update the same files and some unexpected
behaviour might occur.

By using a global mutex, only one publish can be updated at any time.

Fixes #1125

### DIFF
--- a/api/publish.go
+++ b/api/publish.go
@@ -340,6 +340,8 @@ func apiPublishUpdateSwitch(c *gin.Context) {
 	resources = append(resources, string(published.Key()))
 	taskName := fmt.Sprintf("Update published %s (%s): %s", published.SourceKind, strings.Join(updatedComponents, " "), strings.Join(updatedSnapshots, ", "))
 	maybeRunTaskInBackground(c, taskName, resources, func(out aptly.Progress, detail *task.Detail) (*task.ProcessReturnValue, error) {
+		context.PublishMutex.Lock()
+		defer context.PublishMutex.Unlock()
 		err := published.Publish(context.PackagePool(), context, collectionFactory, signer, out, b.ForceOverwrite)
 		if err != nil {
 			return &task.ProcessReturnValue{Code: http.StatusInternalServerError, Value: nil}, fmt.Errorf("unable to update: %s", err)
--- a/context/context.go
+++ b/context/context.go
@@ -40,6 +40,7 @@ type AptlyContext struct {
 	flags, globalFlags *flag.FlagSet
 	configLoaded       bool
 
+	PublishMutex      sync.Mutex
 	progress          aptly.Progress
 	downloader        aptly.Downloader
 	taskList          *task.List
@@ -593,6 +594,7 @@ func NewContext(flags *flag.FlagSet) (*AptlyContext, error) {
 	var err error
 
 	context := &AptlyContext{
+		PublishMutex:      sync.Mutex{},
 		flags:             flags,
 		globalFlags:       flags,
 		dependencyOptions: -1,


dario-gallucci (approved):

randombenj:
After some internal discussion, we noticed that this is actually not an ideal solution. 
Even though it might work, it will block all publish operations for all publish endpoints.
There should be only one lock per publish endpoint.

PR #1135, opened 2022-12-09
Feature/heatlh readiness probe

### LINKED ISSUE #1133: Liveness and Readiness Probe Endpoints
## Detailed Description

It would be nice to have `/ready` and `/healthy` endpoints like described in https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/ for those who deploy to k8s

mmianl:
Fixes #1133

## Requirements

## Checklist

- [x] documentation updated

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 GOVERSION=$(shell go version | awk '{print $$3;}')
 TAG="$(shell git describe --tags --always)"
-VERSION=$(shell echo $(TAG) | sed 's@^v@@' | sed 's@-@+@g')
+VERSION=$(shell echo $(TAG) | sed 's@^v@@' | sed 's@-@+@g' | tr -d '\n')
 PACKAGES=context database deb files gpg http query swift s3 utils
 PYTHON?=python3
 TESTS?=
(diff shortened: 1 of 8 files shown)


mmianl:
Hi,
What do you think of this change?

randombenj:
I definitly like the change! As the `\n` in the version seems to make some problems, what do you think of instead of stripping the newlines in the code (https://github.com/aptly-dev/aptly/pull/1135/files#diff-bf98d5fab5bcfabded9769069e7318683f549ec91cff84c5f87a9ae3a3ca3d8dR28)
We change this in the Makefile: https://github.com/aptly-dev/aptly/blob/master/Makefile#L3

From `VERSION=$(shell echo $(TAG) | sed 's@^v@@' | sed 's@-@+@g')` to `VERSION=$(shell echo $(TAG) | sed 's@^v@@' | sed 's@-@+@g' | tr -d '\n)`

mmianl:
Stripping the newline in the Makefile directly sounds like the better choice to me. I'll update this PR shortly...

mmianl:
Makefile is now changed as suggested and all strings.Trim() calls are removed again.

randombenj (approved):

PR #1139, opened 2022-12-20
Add flag -filter-with-build-deps

### LINKED ISSUE #1131: [Question/Possible PR] Automatically Download Build Dependencies for Source Packages
Hello!

It would be cool if aptly could download the build dependencies of all downloaded source packages automatically.
I tried implementing two approaches to accomplish this and would like to ask for feedback and whether you would be willing to accept a PR for this.

## Detailed Description

We have a script that builds a selection of Debian packages from source for different architectures.
We cache the source packages and their build dependencies using aptly.
Currently, the script computes the build dependencies and then generates a filter query for aptly to download the correct packages.
However, the script could be simplified a lot if aptly had a switch like `-filter-with-build-deps` that works similarly to the existing `-filter-with-deps` but for build dependencies.

## Context

As mentioned above, this change would enable us to simplify a build script.

I think this could be helpful to anyone who is using aptly for caching source packages because when you have a source package, you often also want to have its build dependencies.

## Possible Implementation

I implemented two approaches for achieving this.
The first one is very simple but has a drawback (IMO) but the second one required more code changes.

### Approach 1. Add an Option `-dep-follow-build` Similar to `-dep-follow-source`

This approach exposes the already-existing flag `deb.DepFollowBuild` to the user.
It would basically work like this (note: this is not the complete patch, just the gist of it):
```diff
diff --git a/cmd/cmd.go b/cmd/cmd.go
index 14a0efd1..5cda7ddc 100644
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -114,6 +114,7 @@ package environment to new version.`,
        cmd.Flag.Int("db-open-attempts", 10, "number of attempts to open DB if it's locked by other instance")
        cmd.Flag.Bool("dep-follow-suggests", false, "when processing dependencies, follow Suggests")
        cmd.Flag.Bool("dep-follow-source", false, "when processing dependencies, follow from binary to Source packages")
+       cmd.Flag.Bool("dep-follow-build", false, "when processing dependencies, follow build dependencies")
        cmd.Flag.Bool("dep-follow-recommends", false, "when processing dependencies, follow Recommends")
        cmd.Flag.Bool("dep-follow-all-variants", false, "when processing dependencies, follow a & b if dependency is 'a|b'")
        cmd.Flag.Bool("dep-verbose-resolve", false, "when processing dependencies, print detailed logs")
diff --git a/context/context.go b/context/context.go
index d80528a5..f87e73be 100644
--- a/context/context.go
+++ b/context/context.go
@@ -161,6 +161,9 @@ func (context *AptlyContext) DependencyOptions() int {
                if context.lookupOption(context.config().DepFollowSource, "dep-follow-source") {
                        context.dependencyOptions |= deb.DepFollowSource
                }
+               if context.lookupOption(context.config().DepFollowBuild, "dep-follow-build") {
+                       context.dependencyOptions |= deb.DepFollowBuild
+               }
                if context.lookupOption(context.config().DepVerboseResolve, "dep-verbose-resolve") {
                        context.dependencyOptions |= deb.DepVerboseResolve
                }
```

This is a very simple change. However, it has a disadvantage: When you use `-dep-follow-build` together with `-dep-follow-source`, aptly will now download the sources and build dependencies of your build dependencies (transitively!) instead of just the build dependencies of the original results of the filter.

So if you want the sources + build dependencies of e.g. `bash` you'd have to do
```
... -filter='bash {source}' -dep-follow-build -with-sources -filter-with-deps
```
instead of
```
... -filter='bash' -dep-follow-build -dep-follow-source -with-sources -filter-with-deps
```
Otherwise you get a huge list of packages.

This is why I thought of another approach.

### Approach 2. Add an Option `-filter-with-build-deps` similar to `-filter-with-deps`

In this approach, I added a parameter `withBuildDependencies` to `FilterWithProgress` in `list.go`. 
```go
func (l *PackageList) FilterWithProgress(queries []PackageQuery, withDependencies bool, withBuildDependencies bool, source *PackageList, dependencyOptions int, architecturesList []string, progress aptly.Progress) (*PackageList, error) {
	if !l.indexed {
		panic("list not indexed, can't filter")
	}

	result := NewPackageList()

	for _, query := range queries {
		result.Append(query.Query(l))
	}

	if withDependencies {
		err := addTransitiveDependencies(result, source, dependencyOptions, architecturesList, progress, l)
		if err != nil {
			return nil, err
		}
	}

	if withBuildDependencies {
		// disable DepFollowSource, enable DepFollowBuild
		buildDependencyOptions := dependencyOptions&(^DepFollowSource) | DepFollowBuild
		err := addTransitiveDependencies(result, source, buildDependencyOptions, architecturesList, progress, l)
		if err != nil {
			return nil, err
		}
	}

	return result, nil
}
```

Then I first evaluate the dependencies without the `DepFollowBuild` option as it was before.
Afterwards, I add the build dependencies of all of the source packages but *without* adding their source packages (thus preventing the build dependencies of the build dependencies (and so on) from being added).

## Conclusion

What do you think? Should I open a PR for either of these approaches? Is there maybe a better/smarter way of doing this that I have overlooked?

5hir0kur0:
This flag automatically adds the build dependencies of all source packages matched by the filter expression.

(The sources of the build dependencies and the build dependencies of the build dependencies are *not* added, even if `-dep-follow-source` is specified.)

Resolves #1131 

## Description of the Change

I added a parameter `withBuildDependencies` to `FilterWithProgress` in `list.go`. 
```go
func (l *PackageList) FilterWithProgress(queries []PackageQuery, withDependencies bool, withBuildDependencies bool, source *PackageList, dependencyOptions int, architecturesList []string, progress aptly.Progress) (*PackageList, error) {
	if !l.indexed {
		panic("list not indexed, can't filter")
	}

	result := NewPackageList()

	for _, query := range queries {
		result.Append(query.Query(l))
	}

	if withDependencies {
		err := addTransitiveDependencies(result, source, dependencyOptions, architecturesList, progress, l)
		if err != nil {
			return nil, err
		}
	}

	if withBuildDependencies {
		// disable DepFollowSource, enable DepFollowBuild
		buildDependencyOptions := dependencyOptions&(^DepFollowSource) | DepFollowBuild
		err := addTransitiveDependencies(result, source, buildDependencyOptions, architecturesList, progress, l)
		if err != nil {
			return nil, err
		}
	}

	return result, nil
}
```

Then, I first evaluate the dependencies without the `DepFollowBuild` option as it was before.
Afterwards, I add the build dependencies of all of the source packages but *without* adding their source packages (thus preventing the build dependencies of the build dependencies (and so on) from being added).

## Checklist

  - Happens automatically with `cd man && make generate`. Should I commit the generated `aptly.1` file?
- [x] documentation updated


randombenj:
@5hir0kur0 Sorry for the late response, I will close the MR then

PR #1155, opened 2023-03-07
Fix: Make CodeCov pipeline non-blocking

### LINKED ISSUE #1154: Make CodeCov Pipeline non-blocking
It's nice to have the information that CodeCov provides, but it's annoying to have it block Pull Requests.

## Detailed Description
I'd prefer it if it was more just informational.

## Your Environment
I think this configuration is provided in `codecov.yml`.

reglim:
Fixes #1154 

## Requirements

## Description of the Change
The CodeCov Pipeline will now only fail if the code coverage has fallen more than 2 percent with that Pull Request.

Why this change is important?

--> The pipeline should now no longer fail and block pull requests when only small fixes were applied

## Checklist

- [ ] documentation updated

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -5,3 +5,5 @@ coverage:
         target: auto
         threshold: 0%
         if_ci_failed: error
+        target: auto
+        threshold: 2%


randombenj:
@reglim I think it would be nicer to have something like this:
https://docs.codecov.com/docs/common-recipe-list#set-status-checks-to-block-coverage

Let the code coverage stay at auto and allow a decline of max 2% or so, what do you think?

reglim:
Sounds fair, one moment...

randombenj (approved):

PR #1156, opened 2023-03-09
1125 api failures and race conditions with concurrent publish update operations

### LINKED ISSUE #1125: API: failures and race conditions with concurrent publish update operations

randombenj:
Fixes #1125

## Requirements

## Description of the Change

While implementing the `?_async=true` task api,
resources which must not be handled concurrently
were protected by resource groups.
However the non async resources could still be
operated on concurrently by just doing two api
requests at the same time.

This change fixes this behaviour by
running sync operations in the new task framework
and waiting internally before returning
a result.

## Checklist

- [ ] documentation updated


r4co0n:
You may be aware, but this pull request mixes concerns that seem completely unrelated to me:
* Why we do need to deprecate calls to `rand.Seed` is not explained, and more importantly, I do not see how this is connected to solving API lock errors a.k.a. API race conditions. Why do we need this in here?
* Why should support for multiple Go versions be dropped from the CI (, or generally, following to commit message)?
* Why do we fix a CI linting issue here? Why is this problem seemingly only tracked with a FIXME comment and a link to our dependency's issues?
* Why do we need to bump the Go version here?

Please excuse my rather rude demeanour, I think your changes should probably all be merged, but I see little reasoning as to why this should all be a single pull request.

neolynx:
replaced by #1271

PR #1162, opened 2023-03-13
Snapshot Pull API

reglim:
## Description of the Change

- provide snapshot pull api
- use proper REST api for snapshot merge
  - **this breaks the existing snapshot merge  api, which is only available in CI builds**
- improve swagger doc

## Checklist

- [x] documentation updated
- [x] swagger documented

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,9 +37,8 @@ ifeq ($(RUN_LONG_TESTS), yes)
 endif
 
 install:
-	@echo Building aptly ...
+	@echo "\e[33m\e[1mBuilding aptly ...\e[0m"
 	go generate
-	@echo go install -v
 	@out=`mktemp`; if ! go install -v > $$out 2>&1; then cat $$out; rm -f $$out; echo "\nBuild failed\n"; exit 1; else rm -f $$out; fi
 
 system/env: system/requirements.txt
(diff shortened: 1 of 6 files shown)


neolynx:
codecov/patch is a bit log, but mainly due to a lot of error handling, I think this is acceptable

an0rak-dev (approved):

neolynx:
@cfiehe I updated the swagger doc here, making it look nicer and actually work with the json body params...

2024-10-03-161524_4480x1440_scrot

neolynx:
I discovered that `/snapshots/merge` and `/snapshots/pull` were not implemented as the other /api/snapshot/:name methods and as passing a To to Destination as json arguments. This change also fixes these APIs, leading to a backward incompatibility with the existing `/snapshots/merge`  in CI builds. I hope this does not cause too much troubles...

an0rak-dev:
> This change also fixes these APIs, leading to a backward incompatibility with the existing `/snapshots/merge` in CI builds. I hope this does not cause too much troubles...

@neolynx Maybe we should tag somewhere (the Readme, a commit message, an issue...) that the next release containing this feature should be a major one since backward compatibility is broken with this ?

an0rak-dev (approved):

cfiehe (approved):
Looks good to me, but I must admit, that I am not familiar with using Swagger in Go.

PR #1166, opened 2023-03-17
Api and files

### LINKED ISSUE #975: aptly CLI should be able to work while aptly web is running

### LINKED ISSUE #1017: Serve api and published packages simultaneously

mmianl:
Duplicate of #1116 but with system tests.

Fixes #1017, maybe also fixes #975 but not sure about that

Related documentation change: https://github.com/aptly-dev/www.aptly.info/pull/103

## Requirements

## Checklist

- [x] documentation updated

### DIFF
--- a/api/router.go
+++ b/api/router.go
@@ -51,6 +51,20 @@ func Router(c *ctx.AptlyContext) http.Handler {
 		MetricsCollectorRegistrar.Register(router)
 	}
 
+	router.GET("/repos/:storage/*pkgPath", func(c *gin.Context) {
+		pkgpath := c.Param("pkgPath")
+
+		storage := c.Param("storage")
+		if storage == "-" {
+			storage = ""
+		}
+
+		publicPath := context.GetPublishedStorage("filesystem:" + storage).(aptly.FileSystemPublishedStorage).PublicPath()
+
+		c.FileFromFS(pkgpath, http.Dir(publicPath))
+	})
+
+	api := router.Group("/api")
 	if context.Flags().Lookup("no-lock").Value.Get().(bool) {
 		// We use a goroutine to count the number of
 		// concurrent requests. When no more requests are
@@ -59,7 +73,7 @@ func Router(c *ctx.AptlyContext) http.Handler {
 
 		go acquireDatabase()
 
-		router.Use(func(c *gin.Context) {
+		api.Use(func(c *gin.Context) {
 			var err error
 
 			errCh := make(chan error)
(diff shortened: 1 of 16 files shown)


aol-nnov:
image

Seems, I can make some fuss in the mainline. Should I? :-D

mmianl:
I think this should be approved by @randombenj ;)

aol-nnov:
Definitely! )) But technically, for some reason I have the approve button too here and it reads "Merging can be performed automatically with 1 approving review.", soooo.... ))))

mmianl:
This is probably because I cherry-picked your commit to add system tests, so I assume that this makes you a code owner in the scope of this PR.

I might be wrong though...

system/t12_api/publish.py
```diff
@@ -3,6 +3,11 @@
 
 from api_lib import TASK_SUCCEEDED, APITest
 
+try:
+    import requests
+except ImportError:
+    requests = None
+
```
randombenj, line 10: Why is this needed? I think this should be removed

api/router.go
```diff
@@ -51,6 +51,21 @@ func Router(c *ctx.AptlyContext) http.Handler {
 		MetricsCollectorRegistrar.Register(router)
 	}
 
+	if c.Config().ServeInAPIMode {
```
randombenj, line 54: Cool stuff! What do you think if the `/repos/` endpoint would return  a list (as json?) of all 
the published repos?

randombenj (approved):

api/router.go
aol-nnov: I think, if we really want to extend changes scope to that point, it should be not a json response, but a "directory listing"-like html response similar to that generated by `c.FileFromFS` when requested path is a directory.

system/t12_api/publish.py
mmianl: You are right, I forgot to remove this...

api/router.go
mmianl: How would I determine which repos have been published?

aol-nnov: @mmianl something like that: https://github.com/aptly-dev/aptly/blob/77e02bf7a363c78f9c82792a569c8a5a496c1f06/cmd/publish_list.go#L85-L96

But we need to filter out only `filesystem` publishes, if I get the idea right.

randombenj: Agreed with @aol-nnov , directory listing response like would be nice!
But @mmianl if you think this is out of scope I am happy to merge as is/

aol-nnov: @randombenj I could fiddle with it in a separate PR if my time permits and/or @mmianl won't come with a solution in this PR.

Anyway, current changeset provides viable support for simultaneous serving of files and API and could be merged as is.

aol-nnov:
One more thing came to my mind: it would be nice to align http repo paths during `serve` and `api serve`.

Currently former serves from the root and latter serves from the `/repos`

api/router.go
mmianl: @mmianl @aol-nnov could you provide an example output for this?

aol-nnov: @mmianl I've left my laptop in the office. I'll try to arrange some example tomorrow

aol-nnov: @mmianl I've slept with the idea and actually, it is much easier, than listing publishes an filtering them.

All we need is to list publish prefixes. Rest is handled by the `/repos/:storage/*pkgPath` route!

If you grant me write permissions for this PR, I could push a changeset. Meanwhile find a patch for you latest changes:


```patch
diff --git a/api/repos.go b/api/repos.go
index 8a38dbda..11eddb9e 100644
--- a/api/repos.go
+++ b/api/repos.go
@@ -15,30 +15,20 @@ import (
        "github.com/aptly-dev/aptly/task"
        "github.com/aptly-dev/aptly/utils"
        "github.com/gin-gonic/gin"
-       "github.com/rs/zerolog/log"
 )
 
 // GET /repos
-// todo: limit to filesystem repos only
-// todo: change output format
-func reposListInAPIMode(c *gin.Context) {
-       repos := []*deb.PublishedRepo{}
+func reposListInAPIMode(localRepos map[string]utils.FileSystemPublishRoot) gin.HandlerFunc {
 
-       _ = context.NewCollectionFactory().PublishedRepoCollection().ForEach(func(repo *deb.PublishedRepo) error {
-               err := context.NewCollectionFactory().PublishedRepoCollection().LoadComplete(repo, context.NewCollectionFactory())
-               if err != nil {
-                       msg := fmt.Sprintf(
-                               "Error %s found while getting list of published repos (prefix:%s / distribution:%s / component:%s\n).",
-                               err, repo.StoragePrefix(), repo.Distribution, repo.Components())
-                       log.Debug().Msg(msg)
-                       return err
+       return func(c *gin.Context) {
+               res := c.Writer
+               res.WriteString("<html><body><pre>")
+               for publishPrefix := range localRepos {
+                       res.WriteString(fmt.Sprintf("<a href=\"%[1]s/\">%[1]s</a>\n", publishPrefix))
                }
-
-               repos = append(repos, repo)
-               return nil
-       })
-
-       c.JSON(200, repos)
+               res.WriteString("</pre></body></html>")
+               res.Flush()
+       }
 }
 
 // GET /repos/:storage/*pkgPath
diff --git a/api/router.go b/api/router.go
index 1e368192..ea1dc27f 100644
--- a/api/router.go
+++ b/api/router.go
@@ -53,7 +53,7 @@ func Router(c *ctx.AptlyContext) http.Handler {
        }
 
        if c.Config().ServeInAPIMode {
-               router.GET("/repos/", reposListInAPIMode)
+               router.GET("/repos/", reposListInAPIMode(c.Config().FileSystemPublishRoots))
                router.GET("/repos/:storage/*pkgPath", reposServeInAPIMode)
        }
```

mmianl: Updated accordoing to @aol-nnov's feedback. I changed a few details to make the output of `repos` consistent with the `FileFromFS()` output. I also updated the handler for `/repos/:storage/*pkgPath` according to the above suggestion.

@randombenj, @aol-nnov  what do you think?

api/router.go
```diff
@@ -51,6 +52,12 @@ func Router(c *ctx.AptlyContext) http.Handler {
...
```
aol-nnov, line 55: @mmianl neat!

I have a question, though... `c.Config().ServeInAPIMode` on line 55 that you introduced earlier and `context.Config().FileSystemPublishRoots` on line 56 are, actually from the same context. I'd rather use the same variable for tha sake of consistency. (There are more of such cases down the file, I know, but they are out of scope)

Other than that, LGTM!

mmianl: Fair enough

mmianl:
@randombenj feel free to merge this anytime now

api/router.go
randombenj: Love it thanks :tada:

randombenj:
Thank you so much for this change @mmianl and @aol-nnov !

Issue #1117, opened 2022-09-27
'file already exists and is different' when publishing to S3 with KMS encryption


joshdcox:
## Detailed Description

Even after https://github.com/aptly-dev/aptly/pull/924, publishing to an S3 bucket with KMS encryption enabled results in 'file already exists and is different' errors like the following.

```
ERROR: unable to publish: unable to process packages: error putting file to aptly/focal/pool/main/a/a7xpg/a7xpg_0.11.dfsg1-10build3.1_amd64.deb: file already exists and is different: S3: us-east-1:orca-sre-staging-na1-repo-36812/aptly
```

The objects throwing the errors were PUT by Aptly 1.5.0 to S3 with KMS encryption and contain the user-defined `x-amz-meta-md5` metadata with the correct MD5 checksum. The ETAG of the objects have a length of 32, which is the condition Aptly currently uses to assume the ETAG is an MD5 checksum.

[AWS documentation](https://docs.aws.amazon.com/AmazonS3/latest/API/API_Object.html) states KMS encrypted objects do not use MD5 checksums of the content for their ETAG. It also makes no guarantees about the length of such ETAGs.

https://github.com/aptly-dev/aptly/blob/e90ac6767f35e484e00eae53e1aad6cddb615822/s3/public.go#L326

## Possible Implementation

Instead of assuming an ETAG of length 32 is an MD5 checksum of the content, Aptly could do one of the following:
1.  Use object metadata or upload details (Is this multi-part? Is this KMS encrypted?) to determine if the ETAG can be used for md5 checksum.
1. In the S3PublishEndpoints portion of the Aptly config, allow users to specify whether or not to use object metadata rather than ETAGs for checksum values.
1. Always use object metadata for checksum values on S3 (this would have a performance impact).

Option 2 is probably the simplest solution.
## Your Environment

Aptly version 1.5.0
Running aptly on Ubuntu Focal LTS
Publishing Ubuntu Focal LTS mirrors to S3 buckets with KMS encryption

PR #1167, opened 2023-03-22
Check both MD5 locations for S3 KMS support.

Melraidin:
Fixes https://github.com/aptly-dev/aptly/issues/1117

## Requirements

## Description of the Change

If the S3 bucket used to house a repo has KMS encryption enabled then the etag of an object may not match the MD5 of the file. This may cause an incorrect error to be reported stating the file already exists and is different.

A mechanism exists to work around this issue by using the MD5 stored in object metadata. This check doesn't always cover the case where KMS is enabled as the fallback is only used if the etag is not 32 characters long.

This commit changes the fallback mechanism so that it is used in any case where the object's etag is not 32 characters _or_ if the S3 bucket has encryption enabled for new objects by default.

## Checklist

- [x] documentation updated

### DIFF
--- a/s3/public.go
+++ b/s3/public.go
@@ -53,6 +53,9 @@ type PublishedStorage struct {
 	plusWorkaround   bool
 	disableMultiDel  bool
 	pathCache        map[string]string
+
+	// True if the bucket encrypts objects by default.
+	encryptByDefault bool
 }
 
 // Check interface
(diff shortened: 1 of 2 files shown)


neolynx:
There was an issue with the pathCache, when using the API. 

did you use aptly via cmd or api ?

neolynx:
Rebased on master, some fixes and updated test: https://github.com/aptly-dev/aptly/pull/1272

neolynx:
abandonned https://github.com/aptly-dev/aptly/pull/1272, continuing here

mfolusiak (approved):

PR #1201, opened 2023-07-24
Fix returncode when deleting a mirror with snapshot

acdn-ndostert:
When trying to delete a mirror that has snapshot and not providing the force option, the API should not return a `500
StatusInternalServerError`.
A `403 StatusForbidden` is more appropriate when the condition is expected by the server.

Fixes #

## Requirements

I was able to run `make modules install` and `make test` with success but I'm not able to run `make system-test` because of missing dependencies (gpg1).

## Description of the Change

When implementing a client for Aptly API the `500` HTTP status code returned does not make it clear that there is an issue with the request.
From https://www.rfc-editor.org/rfc/rfc9110.html#name-500-internal-server-error

> [15.6.1. ](https://www.rfc-editor.org/rfc/rfc9110.html#section-15.6.1)[500 Internal Server Error](https://www.rfc-editor.org/rfc/rfc9110.html#name-500-internal-server-error)
> The 500 (Internal Server Error) status code indicates that the server encountered an unexpected condition that prevented it from fulfilling the request.

The condition preventing the request to be fulfill is not unexpected and is even documented in the response.

## Checklist

<s>- [ ] unit-test added (if change is algorithm)</s>
<s>- [ ] functional test added/updated (if change is functional)</s>
<s>- [ ] man page updated (if applicable)</s>
<s>- [ ] bash completion updated (if applicable)</s>
<s>- [ ] documentation updated</s>

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -54,3 +54,4 @@ List of contributors, in chronological order:
 * Boxjan (https://github.com/boxjan)
 * Mauro Regli (https://github.com/reglim)
 * Alexander Zubarev (https://github.com/strike)
+* Nicolas Dostert (https://github.com/acdn-ndostert)
--- a/api/mirror.go
+++ b/api/mirror.go
@@ -157,7 +157,7 @@ func apiMirrorsDrop(c *gin.Context) {
 			snapshots := snapshotCollection.ByRemoteRepoSource(repo)
 
 			if len(snapshots) > 0 {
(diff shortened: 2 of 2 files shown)


reglim:
It's HTTP `403`. not `404` as you said in the description, but I agree that StatusForbidden makes more sense.

reglim (changes requested):
Can you change 404 to 403 in the commit and PR description? Otherwise LGTM

PR #1208, opened 2023-08-17
Fix: Remove Batch from API options, set to true by default

reglim:
It doesn't make sense to expose the Batch option, which, if `false` enables input from stdin for gpg passphrases, which blocks the api thread. Because of this, I removed the option, and set it to `true` by default.

## Requirements

## Description of the Change

* Remove `Batch` option
* Set `Batch` to true internally

Why this change is important?

So the API can't get hung up because a user tries to sign a publish with a key that requires a passphrase he didn't provide in the request.

## Checklist

- [x] documentation updated

### DIFF
--- a/api/publish.go
+++ b/api/publish.go
@@ -16,7 +16,6 @@ import (
 // SigningOptions is a shared between publish API GPG options structure
 type SigningOptions struct {
 	Skip           bool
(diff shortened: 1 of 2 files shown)


randombenj (approved):
LGTM

Issue #1213, opened 2023-08-25
aptly graph image is removed immediately after rendering


artem-kamyshev:
I try to view a graph. I see it opened in Gwenview for a moment, but then the viewer closes it and shows a tmp folder and there's no aplty-generated files in it which I can confirm with `ls` for example. The file is generated but it's removed in a second or two for some reason.

## Detailed Description
akamyshev@workpc:~/repostests$ /opt/aptly/aptly_1.5.0_linux_amd64/aptly graph
Generating graph...
Rendered to png file: /tmp/aptly-graph4049935713.png, trying to open it with: xdg-open /tmp/aptly-graph4049935713.png...
org.kde.kdegraphics.gwenview.lib: Unresolved mime type  "image/x-mng"
org.kde.kdegraphics.gwenview.lib: Unresolved raw mime type  "image/x-nikon-nrw"
org.kde.kdegraphics.gwenview.lib: Unresolved raw mime type  "image/x-samsung-srw"
org.kde.kdegraphics.gwenview.app: Found no root index for QUrl("file:///tmp")
akamyshev@workpc:~/repostests$ org.kde.kdegraphics.gwenview.lib: Could not generate thumbnail for file "file:///tmp/aptly-graph4049935713.png"

## Your Environment
Debian 11
aptly 1.5.0

reglim:
Just tried it and had the same issue, I'll look at it.

PR #1214, opened 2023-09-13
Fix: Graph deleted before aptly exits

reglim:
Fixes: #1213

## Description of the Change

A viewer might reload/monitor the image file, therefore aptly should not remove the temporary image file.

As the tempfile is created in /tmp, it should be cleanedup automatically or by the user.


randombenj (approved):

iofq (approved):

neolynx:
rewrite of original PR, instead of a timeout and removing the tmp file, just leave it to the user or system to clean it up.

szakalboss (approved):

PR #1217, opened 2023-09-13
Chore: 1216-update-go-python-and-dependencies

### LINKED ISSUE #1216: Update Golang to 1.21, Python to 3.11 and Dependencies
It's been a while : )

[other questions removed because it should be self-explanatory]

reglim:
Fixes #1216

## Requirements

## Description of the Change

* Update Go to 1.21 in `go.mod`
* Remove matrix from CI, run Pipeline with 1.21
* Update `setup-python` in CI from v2 to v4
* Update CI to use Python 3.11 instead of 3.9
* Implement linting suggestions from updated version of golangci-lint to make pipeline pass
* Change expected output of `IncludeRepo3Test`, as the error message comes from go itself, which seems to have changed since version 1.16, see code below:

```go
var repoTemplate *template.Template
repoTemplate, err = template.New("repo").Parse(repoTemplateString)
if err != nil {
	return fmt.Errorf("error parsing -repo template: %s", err)
}
```

## Why this change is important?

The dependencies have not been updated for quite some time, and Go was on version 1.16 which is now two years old, so it was time for an upgrade.

## Checklist

- [ ] documentation updated


swift/public_test.go
```diff
@@ -27,8 +26,6 @@ var _ = Suite(&PublishedStorageSuite{})
 func (s *PublishedStorageSuite) SetUpTest(c *C) {
 	var err error
 
-	rand.Seed(int64(time.Now().Nanosecond()))
```
randombenj, line 30: Might this have some implications, why was this deprecated?

reglim: It should be seeded automatically at startup:

https://github.com/golang/go/issues/56319

randombenj (approved):
LGTM thanks, feel free to merge when ready

PR #1219, opened 2023-09-15
Shut down cleanly when 'api serve' is interrupted

refi64:
This will properly close the db and, more particularly, flush out any profile files being written. Otherwise, they can end up empty or truncated.


refi64:
Pushed a slightly cleaned up version.

randombenj (approved):

PR #1220, opened 2023-09-15
Improve test output regexes for better perf and Go 1.20 support 

refi64:
From the first commit:

```
The current regex runs in exponential time, which massively impacts the
runtime of the test suite, taking several seconds (~4s on my system)
just to perform a single match. By replacing the mix of re.findall + the
initial capture group with re.search + some string slicing, the time
spent matching the regex becomes nearly instant, e.g.:

    $ make system-test TESTS='Config*'

goes from taking ~10s to ~1.5s.
```

## Checklist

- [ ] documentation updated


reglim:
LGTM, @randombenj what do you think?

Also, would you mind adding your name to the `AUTHOR` file?

refi64:
@reglim added!

reglim (approved):

PR #1228, opened 2023-10-31
replace `golang.org/x/crypto/openpgp` with `github.com/ProtonMail/go-crypto/openpgp`

paulcacheux:
Fixes some TODOs in the code

## Requirements

## Description of the Change

This PR replaces `golang.org/x/crypto/openpgp` which is unsupported ([issue](https://github.com/golang/go/issues/44226)) with `github.com/ProtonMail/go-crypto/openpgp`.

This has the added bonus of adding support for new pgp key types

## Checklist

- [ ] documentation updated


lebauce (approved):

paulcacheux:
@randombenj sorry for the direct ping. Is this something you would be interested in ?

pgp/openpgp.go
```diff
@@ -94,12 +92,6 @@ func checkDetachedSignature(keyring openpgp.KeyRing, signed, signature io.Reader
 			sigType = sig.SigType
 			creationTime = sig.CreationTime
 			pubKeyAlgo = sig.PubKeyAlgo
-		case *packet.SignatureV3:
```
randombenj, line 97: Will this have any impact?

randombenj (commented):
Thanks for the change, I think this makes sense. Why was the V3 switch removed?

pgp/openpgp.go
paulcacheux: Yes something I should have made clearer in the PR description. Basically this means we would lose support for SignatureV3. They are old and nearly unsupported so I think it's your decision.

On the other hand, the current openpgp does not support ed25519 keys ([example issue](https://github.com/gopasspw/gopass/issues/2192)), and is not supported. So migrating has both upsides and downsides.

I searched for another maintained `/x/crypto/openpgp` alternatives that is still maintained without success sadly

paulcacheux: After double checking https://github.com/ProtonMail/go-crypto/blob/afb1ddc0824ce0052d72ac0d6917f144a1207424/openpgp/packet/packet.go#L339, I think there was already now way to read a SignatureV3 from a packet, so I think we don't lose anything.

Here is a public issue showing a similar intel https://github.com/golang/go/issues/27679

paulcacheux:
Regarding the failing tests it seems the upstream repos were updated. Happy to rebase once main is green regarding those

pgp/openpgp.go
randombenj: @neolynx any  input on this, otherwise I think we can merge

paulcacheux:
@randombenj maybe we can fix the CI first before merging. How can I help in that direction ?

pgp/openpgp.go
neolynx: I would go with that. We can test with old mirrors and repos, and readd it in case this brings any troubles...

neolynx (approved):
lgtm :)

randombenj:
@paulcacheux as the tests don't fail because of your changes, I will merge this and then later on fix the tests on the main branch. Thanks for the contribution :tada:

PR #1231, opened 2023-11-08
fix failing build on hosts with wildcard dns (fixes #1230)

### LINKED ISSUE #1230: nosuch.host tests fail when building on a host where wildcard dns entries are available
When building on a host which has whose DNS search list contains a wildcard domain, tests fail

## Detailed Description
Assume a host where the local search domain contains a wildcard alias `*.example`, then `nosuch.host`
will be resolved as `nosuch.host.example` which then very well exists.

As `host` may become a valid TLD some day, it might be clever anyways, to switch to `.invalid.` TLD as recommended by RFC2606.

## Context

Should there be another user in a similar environment, he couldn't successfully test the code basis although the error
is not to be located in the code but the build environment.

## Possible Implementation

I'm already preparing a pull request

## Your Environment

`nosuch.host` (without a dot at the end) actually resolves!

Jennes:
on hosts which have wildcard dns domains in their local domain search list, builds failed because "nosuch.host" could actually be resolved.

Since ".host" isn't a recommended TLD by RFC2606, we use ".invalid" now. And since this is not enough to fix the problem, we use now absoulte domain names (having a '.' at the end)

Fixes #

## Requirements

## Checklist

- [ ] documentation updated

### DIFF
--- a/http/download_test.go
+++ b/http/download_test.go
@@ -115,7 +115,7 @@ func (s *DownloaderSuite) TestDownload404(c *C) {
 }
 
 func (s *DownloaderSuite) TestDownloadConnectError(c *C) {
-	c.Assert(s.d.Download(s.ctx, "http://nosuch.host/", s.tempfile.Name()),
+	c.Assert(s.d.Download(s.ctx, "http://nosuch.host.invalid./", s.tempfile.Name()),
 		ErrorMatches, ".*no such host")
 }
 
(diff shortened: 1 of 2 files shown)


Jennes:
On my local machines the tests work fine (with or without the wildcard domains). So I guess there is something wrong with DNS resolution on the Github runner.
But what would be a decent test which also works within Github?

neolynx:
This looks to me like a testing error. how being able to resolve the host, means download failed. so the test should be successful ?

Issue #287, opened 2015-08-21
Snapshots diff sometimes returns null side with older package


fpytloun:
I have two snapshots, older one (`_tcp-1440157930`) and newer (`_tcp-1440161185`). Only difference between them is that newer snapshot contains package `python-aptly 0.5.2`

When I compare these snapshots, I always get `null` on the side of older snapshot:

```
$ curl http://localhost:8080/api/snapshots/_tcp-1440157930/diff/_tcp-1440161185
[{"Left":null,"Right":"Pall python-aptly 0.5.2 c0f4a0c3fc9e22c5"}]
$ curl http://localhost:8080/api/snapshots/_tcp-1440161185/diff/_tcp-1440157930
[{"Left":"Pall python-aptly 0.5.2 c0f4a0c3fc9e22c5","Right":null}]
```

So it seems that version comparison like `0.5.2 > 0.5` doesn't work.
Expected output would be:

```
[{"Left":"Pall python-aptly 0.5 f0dfd70e4a936c2f","Right":"Pall python-aptly 0.5.2 c0f4a0c3fc9e22c5"}]
```

This is how newer snapshot looks like:

```
$ curl -sss http://localhost:8080/api/snapshots/_tcp-1440161185/packages|json_pp|grep python-aptly|sort
   "Pall python-aptly 0.2.2 cbd9cf268ece7eb2",
   "Pall python-aptly 0.3.1 8142186f92d78d20",
   "Pall python-aptly 0.4.1 950bac934669ffac",
   "Pall python-aptly 0.4.2 984dd9ec9b30e4bf",
   "Pall python-aptly 0.4.3 dc6e6adee37a0486",
   "Pall python-aptly 0.4.4 d943dbd2a4bf39c9",
   "Pall python-aptly 0.4.6 98288a131f1b253b",
   "Pall python-aptly 0.4 864efe6993977f7f",
   "Pall python-aptly 0.5.2 c0f4a0c3fc9e22c5",
   "Pall python-aptly 0.5 f0dfd70e4a936c2f",
```

And the older one:

```
$ curl -sss http://localhost:8080/api/snapshots/_tcp-1440157930/packages|json_pp|grep python-aptly|sort
   "Pall python-aptly 0.2.2 cbd9cf268ece7eb2",
   "Pall python-aptly 0.3.1 8142186f92d78d20",
   "Pall python-aptly 0.4.1 950bac934669ffac",
[...]

PR #1233, opened 2023-11-29
Fix reflist diffs failing to compact when one of the inputs ends

### LINKED ISSUE #282: aptly diff misses some packages
Hi!

Here is an output of aptly diff:

```
$ aptly snapshot diff extern-datastax-datastax-main-2015-08-12-1 extern-datastax-datastax-main-2015-07-01-1
  Arch   | Package                                  | Version in A                             | Version in B
- all    | datastax-agent                           | 5.2.0                                    | -       
- all    | opscenter                                | 5.2.0                                    | -       
```

For some reasons, no version is seen in "B". But, they are here:

```
$ aptly snapshot show -with-packages extern-datastax-datastax-main-2015-07-01-1
Name: extern-datastax-datastax-main-2015-07-01-1
Created At: 2015-07-01 11:19:18 CEST
Description: Snapshot from mirror [extern-datastax-datastax-main]: http://debian.datastax.com/community/ stable
Number of packages: 31
Packages:
  datastax-agent_4.0.0_all
  datastax-agent_4.0.1_all
  datastax-agent_4.0.2_all
  datastax-agent_4.0.3_all
  datastax-agent_4.1.0_all
  datastax-agent_4.1.1_all
  datastax-agent_4.1.2_all
  datastax-agent_4.1.3_all
  datastax-agent_4.1.4_all
  datastax-agent_5.0.0_all
  datastax-agent_5.0.1_all
  datastax-agent_5.0.2_all
  datastax-agent_5.1.0_all
  datastax-agent_5.1.1_all
  datastax-agent_5.1.2_all
  datastax-agent_5.1.3_all
  opscenter_4.0.1_all
  opscenter_4.0.2_all
  opscenter_4.0.3_all
  opscenter_4.1.0_all
  opscenter_4.1.1_all
  opscenter_4.1.2_all
  opscenter_4.1.3_all
  opscenter_4.1.4_all
  opscenter_5.0.0_all
  opscenter_5.0.1_all
  opscenter_5.0.2_all
  opscenter_5.1.0_all
  opscenter_5.1.1_all
  opscenter_5.1.2_all
  opscenter_5.1.3_all
```

Maybe aptly is confused by the fact that the package is present in different version. I have the problem with other repositories with the same characteristic, but not on all of them. For example:

```
$ aptly snapshot diff extern-jenkins-jenkins-2015-08-12-1 extern-jenkins-jenkins-2015-06-26-2
  Arch   | Package                                  | Version in A                             | Version in B
+ all    | jenkins                                  | -                                        | 1.617   
- all    | jenkins                                  | 1.624                                    | -       
```

refi64:
The previous reflist logic would early-exit the loop body if one of the lists was empty, but that skips the compacting logic entirely.

Instead of doing the early-exit, we can leave a list's ref as nil when the list end is reached and then flip the comparison result, which will essentially treat it as being greater than all others. This should preserve the general behavior without omitting the compaction.

<hr>

I *suspect* this might fix #282 and/or #287, but I haven't tested that.

An alternate approach in the same vein is to create a "sentinel" slice like:

```go
sentinel := []byte{255}
```

and then use that when we reach the end of a reflist, which avoids the need to flip the comparison result. However, the code later on that checks if the ref is set would then look like:

```go
if pl == nil && rl[0] != sentinel[0]
```

which felt worse to me while also being a tad wasteful (extra slice allocation every call!). That being said, omitting the comparison flip could certainly be seen as cleaner.

## Checklist

- [ ] documentation updated

### DIFF
--- a/deb/reflist.go
+++ b/deb/reflist.go
@@ -194,31 +194,21 @@ func (l *PackageRefList) Diff(r *PackageRefList, packageCollection *PackageColle
 
 	// until we reached end of both lists
 	for il < ll || ir < lr {
-		// if we've exhausted left list, pull the rest from the right
-		if il == ll {
-			pr, err = packageCollection.ByKey(r.Refs[ir])
-			if err != nil {
-				return nil, err
-			}
-			result = append(result, PackageDiff{Left: nil, Right: pr})
-			ir++
-			continue
+		var rl, rr []byte
+		if il < ll {
+			rl = l.Refs[il]
 		}
-		// if we've exhausted right list, pull the rest from the left
-		if ir == lr {
-			pl, err = packageCollection.ByKey(l.Refs[il])
-			if err != nil {
-				return nil, err
-			}
-			result = append(result, PackageDiff{Left: pl, Right: nil})
-			il++
-			continue
+		if ir < lr {
+			rr = r.Refs[ir]
 		}
 
-		// refs on both sides are present, load them
-		rl, rr := l.Refs[il], r.Refs[ir]
 		// compare refs
 		rel := bytes.Compare(rl, rr)
+		// an unset ref is less than all others, but since it represents the end
+		// of a reflist, it should be *greater*, so flip the comparison result
+		if rl == nil || rr == nil {
+			rel = -rel
+		}
 
 		if rel == 0 {
 			// refs are identical, so are packages, advance pointer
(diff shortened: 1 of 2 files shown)


refi64:
CI failures seem to be unrelated? Looks like `r-project` updated their packages:

```diff
-Success downloading https://cloud.r-project.org/bin/linux/debian/bullseye-cran40/python3-rpy2-dbgsym_3.5.10-1~bullseyecran.0_amd64.deb
-Success downloading https://cloud.r-project.org/bin/linux/debian/bullseye-cran40/python3-rpy2-dbgsym_3.5.10-1~bullseyecran.0_i386.deb
+Success downloading https://cloud.r-project.org/bin/linux/debian/bullseye-cran40/littler_0.3.18-2~bullseyecran.0_all.deb
+Success downloading https://cloud.r-project.org/bin/linux/debian/bullseye-cran40/python3-rpy2-dbgsym_3.5.12-1~bullseyecran.0_amd64.deb
```

PR #1222, opened 2023-09-25
Improve publish performance, especially for prefixes with a large number of snapshots

refi64:
This contains a variety of improvements for publish performance, specifically speeding up the cleanup operation by:

- using a separate, faster package to walk the filesystem
- avoiding re-loading the same packages repeatedly (this will happen if a single prefix has a large number of snapshots or repos that share most of their packages)
- slightly improving the package list loading performance w/ zero-copy deserialization

Benchmarks were added for all of these, and some unit tests were added specifically to test aspects of cleanup.

We have a relatively large aptly repository with >90 repositories, ~207k packages across all of the repositories, and >3.5k snapshots; a testing version of that repository was used to measure the publishing performance. Prior to these changes, publishing took >9 *minutes*, with over 8 minutes of that time just in the cleanup phase. With these, the cleanup time goes down to ~13 *seconds*, for a total publish time of a little under a minute.

## Checklist

- [x] unit-test added (if change is algorithm)
- [ ] functional test added/updated (if change is functional)
- [ ] man page updated (if applicable)
- [ ] bash completion updated (if applicable)
- [ ] documentation updated
- [ ] author name in `AUTHORS` (I already have a commit for this in #1220 and don't want merge conflicts)

PR #1235, opened 2023-12-07
[RFC] Split reflists to share their contents across snapshots

refi64:
This builds on top of #1222, #1227, and #1233, and is thus in draft state until those are merged in. The only commit that's actually new is the [very last one](https://github.com/aptly-dev/aptly/commit/0d798671bf2c8cbe02b2ab95fbc46fc84b896299), whose commit message I copied below.  (Even that single commit is admittedly quite big, but a sizable chunk of the changes are just plumbing a new `RefListCollection` around across the code.)

## Description of the Change

In current aptly, each repository and snapshot has its own reflist in
the database. This brings a few problems with it:

- Given a sufficiently large repositories and snapshots, these lists can
  get enormous, reaching >1MB. This is a problem for LevelDB's overall
  performance, as it tends to prefer values around the confiruged block
  size (defaults to just 4KiB).
- When you take these large repositories and snapshot them, you have a
  full, new copy of the reflist, even if only a few packages changed.
  This means that having a lot of snapshots with a few changes causes
  the database to basically be full of largely duplicate reflists.
- All the duplication also means that many of the same refs are being
  loaded repeatedly, which can cause some slowdown but, more notably,
  eats up huge amounts of memory.
- Adding on more and more new repositories and snapshots will cause the
  time and memory spent on things like cleanup and publishing to grow
  roughly linearly.

*Split reflists* aim at solving this by separating reflists into 64
*buckets*. Package refs are sorted into individual buckets according to
the following system:

- Take the first 3 letters of the package name, after dropping a `lib`
  prefix. (Using only the first 3 letters will cause packages with
  similar prefixes to end up in the same bucket, under the assumption
  that packages with similar names tend to be updated together.)
- Take the 64-bit xxhash of these letters. (xxhash was chosen because it
  relatively good distribution across the individual bits, which is
  important for the next step.)
- Use the first 6 bits of the hash (range [0:63]) as an index into the
  buckets.

Once refs are placed in buckets, a sha256 digest of all the refs in the
bucket is taken. These buckets are then stored in the database, split
into roughly block-sized segments, and all the repositories and
snapshots simply store an array of bucket digests.

This approach means that *repositories and snapshots can share their
reflist buckets*. If a snapshot is taken of a repository, it will have
the same contents, so its split reflist will point to the same buckets
as the base repository, and only one copy of each bucket is stored in
the database. When some packages in the repository change, only the
buckets containing those packages will be modified; all the other
buckets will remain unchanged, and thus their contents will still be
shared. Later on, when these reflists are loaded, each bucket is only
loaded once, short-cutting loaded many megabytes of data. In effect,
split reflists are essentially copy-on-write, with only the changed
buckets stored individually.

Changing the disk format means that a migration needs to take place, so
that task is moved into the database cleanup step, which will migrate
reflists over to split reflists, as well as delete any unused reflist
buckets.

All the reflist tests are also changed to additionally test out split
reflists; although the internal logic is all shared (since buckets are,
themselves, just normal reflists), some special additions are needed to
have native versions of the various reflist helper methods.

- Memory usage during publish and database cleanup, with
  `GOMEMLIMIT=2GiB`, goes down from ~3.2GiB (larger than the memory
  limit!) to ~0.7GiB, a decrease of ~4.5x.

*In my local tests*, publish times had also decreased down to mere
seconds but the same effect wasn't observed on the server, with the
times staying around the same. My suspicions are that this is due to I/O
performance: my local system is an M1 MBP, which almost certainly has
much faster disk speeds than our DigitalOcean block volumes. Split
reflists include a side effect of requiring more random accesses from
reading all the buckets by their keys, so if your random I/O
performance is slower, it might cancel out the benefits. That being
said, even in that case, the memory usage and database size advantages
still persist.

## Checklist

- [ ] documentation updated

<hr>

It would be awesome if anyone could also test this out and report how it affects their performance & memory usage.


neolynx:
Thanks for the analysis and solution ! I will give it a try on my side...

neolynx:
Could you fix the golang-ci-lint errors ? 

You can see them by running: `docker run --rm -v $(pwd):/app -w /app golangci/golangci-lint golangci-lint run -v`

Then the pipeline passes further :)

refi64:
@neolynx whoops didn't notice those, fixed now!

neolynx:
Now we have some tests failing:

https://github.com/aptly-dev/aptly/actions/runs/7558329700/job/20579601126?pr=1235#step:10:5192

not sure this is related... could you maybe check what this is ?

refi64:
ah I believe those are because the source apt repos themselves changed, so now the mirrored packages also changed. I think I've seen that test fixed before via manual updates...not sure if you'd like something like that in this PR or elsewhere.

refi64:
@neolynx I fixed the mirror tests separately in #1244. I did also discover a small bug buried in the other mirror tests (!) and fixed it here, but I left the commits in #1244 separate to avoid merge hell.

neolynx:
working on #1222 in #1273

neolynx:
replaced by #1282

PR #1245, opened 2024-02-04
Fix/system tests

neolynx:
Fixes the system tests permanently, by using a mirror on S3 for all the required files for the tests. No direct internet download, no more flaky tests.

Additionally:
- use ubuntu 22.04 runner in gitlab ci
- switch to gpg2
- add a system test docker image, and make targets


randombenj (approved):
Awesome work! @neolynx This finally makes the tests really reproducible and not depending on internet resources!

PR #1246, opened 2024-02-06
gpg api: allow self signed and use default gpg version

neolynx:
## Description of the Change

Self signed GPG keys were refused, which is something aptly should allow, if a repo for mirroring is singed this way.

Also, use the default GPG version, as in the other parts for using GPG keys.

The API also returns the gpg command output, instead of just a failure.

## Checklist

- [ ] documentation updated

### DIFF
--- a/api/gpg.go
+++ b/api/gpg.go
@@ -25,7 +25,7 @@ func apiGPGAddKey(c *gin.Context) {
 	}
 
 	var err error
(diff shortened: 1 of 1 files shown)


mfolusiak (approved):
`and use default gpg version` is probably redundant in the commit message, otherwise works as expected.

PR #1253, opened 2024-03-10
Use tagged version of ProtonMail/go-crypto in go.mod

paulcacheux:
## Description of the Change

This change convert the usage of `github.com/ProtonMail/go-crypto` from a commit based require to a tag one. The underlying commit is currently exactly the same one.

When we started using ProtonMail/go-crypto the project did not use tag for releases which forced us to use the commit in the go.mod file. Now that they are tagging releases, we can move on to the actual tag.

## Checklist

- [ ] documentation updated


neolynx:
looks good to me. could you add a description why this is needed ?

PR #1264, opened 2024-04-10
source package upload: allow + in version

neolynx:
## Requirements

## Description of the Change

Aptly ignores source package archive files if they have a plus in the version. This happens because files uploaded to aptly have plus signes url escaped, but the filename inside the dsc file is not escaped. This changes fixes this behavior, by unescaping the filename when needed.

## Checklist

- [ ] documentation updated


neolynx:
this error is caused when a client urlencodes the filename in the multipart form. not a aptly bug.

PR #1270, opened 2024-04-11
github CI: nightly builds for multiple distributions

### LINKED ISSUE #1251: Nightly version for Debian/Ubuntu requires GLIBC_2.32 since the latest version
I am using the nightly repository of aptly `deb http://repo.aptly.info/ nightly main` and noticed, that the current version is not running on Ubuntu Focal 20.04 anymore:

```
root@a49b79157c31:/# aptly
aptly: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by aptly)
aptly: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by aptly)
``` 

## Detailed Description

I guess the issue is caused by a change in the build environment. Probably the current version got build on a newer version of Linux, such as Ubuntu 22.04 or Debian 12 for example.

Ubuntu 20.04 on the other hand uses `GLIBC 2.31`. The latest known nightly build that is running on GLIB 2.31 systems is `1.5.0+64+gf1649a64` which can be installed via `apt-get install aptly=1.5.0+64+gf1649a64`

## Context

I know that the nightly repository is not for production and the stable repository still ships a version that runs fine on Ubuntu 20.04. With that said, I guess there is no benefit in requiring  `GLIBC_2.32` and the issue is just due to the fact, that the current version of aptly is getting build on a more modern system.

Ubuntu 20.04 has still active support until Apr 2025, having still support for it would be nice.

## Possible Implementation
To be as compatible as possible, I would recommend to use an older version of Debian/Ubuntu to build aptly. I personally use Docker for this. For example the official golang docker image `1.22.0-bullseye` should resolve this issue. 

## Your Environment

Ubuntu 20.04, amd64

I'm using the nightly version because there where some missing features in the stable, cant remember which feature it was exactly. I know this is an edge case, noting to important.

neolynx:
Fixes #1251

## Description of the Change

    Since the pipeline changed to use ubuntu22.04 runners, the nighty debian packages do not work on debian buster anymore.
    This change updates the pipeline to build for Ubuntu 20.04 and 22.04 as well as for
    Debian 10, 11 and 12.
    
    The distribution specific apt sources are as follows:

      "deb http://repo.aptly.info/nightly-bookworm bookworm main"

    (replace bookworm with focal, buster, bullseye. Install aptly repo key with: curl -fsS https://www.aptly.info/pubkey.txt | sudo gpg --dearmor -o /etc/apt/trusted.gpg.d/aptly.gpg)

    The builds on focal will also release to the legacy nightly apt repo: https://github.com/aptly-dev/aptly/actions/runs/8723898496/job/23933824692#step:7:24

    This only affects nightly builds, for now.
    Pipeline example: [](https://github.com/aptly-dev/aptly/actions/runs/8723898496)


upload-artifacts.sh
```diff
@@ -10,50 +10,100 @@ aptly_password="$APTLY_PASSWORD"
...
+    done
     echo
-done
+}
+cleanup() {
+    echo "\nCleanup..."
+    curl -fsS -X DELETE  -u $aptly_user:$aptly_password ${aptly_api}/api/files/$folder
+    echo
+}
+trap cleanup EXIT
```
szakalboss, line 47: wanna trap also INT and TERM ?

szakalboss (approved):

upload-artifacts.sh
neolynx: I think both result in EXIT

PR #1281, opened 2024-04-20
fix:failed to edit repo

neolynx:
return when  the repo cant be found and err !=nil

Replaces #1093

## Requirements

## Checklist

- [ ] documentation updated

### DIFF
--- a/api/repos.go
+++ b/api/repos.go
@@ -115,8 +115,8 @@ func apiReposEdit(c *gin.Context) {
 
 	if b.Name != nil {
 		_, err := collection.ByName(*b.Name)
-		if err == nil {
-			// already exists
+		if err != nil {
+			// repo does not exist
 			AbortWithJSONError(c, 404, err)
 			return
 		}


api/repos.go
```diff
@@ -115,8 +115,8 @@ func apiReposEdit(c *gin.Context) {
 
 	if b.Name != nil {
 		_, err := collection.ByName(*b.Name)
-		if err == nil {
-			// already exists
+		if err != nil {
+			// repo does not exist
```
mfolusiak, line 119: Some description to the PR is needed. What is the intention here?
Why do we raise error if  `*b.Name` doesn't exist?  If we want to handle renaming of the repo from `:name` to `*b.Name` the second one should not exist.

mfolusiak: I just saw similar doubt has been expressed by @randombenj in the original PR https://github.com/aptly-dev/aptly/pull/1093#issuecomment-1207839336
I suggest to close both PRs without merging if this comment is not addressed

neolynx: to me, this looks like as follows:
- edit repo "A" and set name to "B"
- if no repo "A" exists, abort (line 111)
- if new name is set (line 116)
- check if repo with new "B" name already exists (line 117)
- if yes (line 118: if err == nil), abort, because we cannot rename to an existing repo

therefore, the patch makes no sense, it incorrectly aborts when getting the repo "B" does return an error (err != nil) and the "repo does not exist.

I agree to drop the PR.

PR #1292, opened 2024-05-18
Introduce option `-multi-dist` to the publish commands

### LINKED ISSUE #486: Impossible to publish multiples distributions with same prefix
It appears that it is impossible to publish more than one distribution on with same prefix.

## Detailed Description
For my needs, I have to mirror two distributions of the same archive and publish them.

I can publish the first distribution without problem (thanks to a snapshot of the mirror). However, when the second distribution is published, the following error appears (for example) :

> ERROR: unable to publish: unable to process packages: error linking file to /root/.aptly/public/hardware-mcp/pool/main/s/sblim-cmpi-devel/libcmpicppimpl0_2.0.3-0ubuntu2_amd64.deb: **file already exists and is different**

After verification, my two distributions contain packages unchanged from one to the other (same versions but **different checksums**).

A workaround I could find is to include the distribution name in the prefix but this does not seem very practical.

## Possible Implementation
After some investigations, it seems that this is due to the architecture of files retained for publication and especially that of the folder _pool_ of the published archive.

There come to mind a possible solution: create a subfolder by distribution in the _pool_ folder (by component too ?).

## Your Environment
OS : Debian Jessie 8.7
Aptly : 0.9.7

nresare:
This change makes it possible to publish multiple distributions with packages named the same but with different content by changing structure of the generated pool hierarchy. The option not enabled by default as this changes the structure of the output which could break the expectations of other tools.

Fixes #486 

## Requirements

## Description of the Change

Given #486 and and [this discussion on aptly-discuss](https://groups.google.com/g/aptly-discuss/c/YTLXLUqBbFo) it seems I am not alone in using aptly to self-publish some packages for a handful different distributions and wanting to publish different packages with the same filename for different distributions. As it stands, due to aptly using the default directory structure for the `pool` hierarchy, aptly will currently fail to publish such packages.

A casual look at  the [DebianRepository/Format page](https://wiki.debian.org/DebianRepository/Format) seems to indicate that the structure of the pool hierarchy is not specified, and as long as the `Filename` field in Packages references the relevant file any structure could be used.

This change introduces a new option `-multi-dist` that will create per distribution subdirectories under `pool` such that each distribution can have its own copy of some package and need not worry about two packages with the same name but different distributions colliding.

One could argue that it would be better to change the default behaviour, rather than implementing a new option, but I feel I have limited visibility into all the implications here, so having this behaviour be enabled on an opt-in basis seems the choice with less risk of functionality regressions. This change can also have storage efficiency implications, so having it be optional seems preferable for that reason as well.

## Checklist

- [ ] documentation updated

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -59,3 +59,4 @@ List of contributors, in chronological order:
 * Paul Cacheux (https://github.com/paulcacheux)
 * Nic Waller (https://github.com/sf-nwaller)
 * iofq (https://github.com/iofq)
+* Noa Resare (https://github.com/nresare)
--- a/api/publish.go
+++ b/api/publish.go
@@ -101,6 +101,7 @@ func apiPublishRepoOrSnapshot(c *gin.Context) {
 		Architectures        []string
 		Signing              SigningOptions
 		AcquireByHash        *bool
+		MultiDist            bool
 	}
 
 	if c.Bind(&b) != nil {
@@ -226,7 +227,7 @@ func apiPublishRepoOrSnapshot(c *gin.Context) {
 			return &task.ProcessReturnValue{Code: http.StatusBadRequest, Value: nil}, fmt.Errorf("prefix/distribution already used by another published repo: %s", duplicate)
 		}
 
-		err := published.Publish(context.PackagePool(), context, collectionFactory, signer, publishOutput, b.ForceOverwrite)
+		err := published.Publish(context.PackagePool(), context, collectionFactory, signer, publishOutput, b.ForceOverwrite, b.MultiDist)
 		if err != nil {
 			return &task.ProcessReturnValue{Code: http.StatusInternalServerError, Value: nil}, fmt.Errorf("unable to publish: %s", err)
 		}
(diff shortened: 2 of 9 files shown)


neolynx:
looks great ! thank you so much :+1: 

would you mind adding a PR to https://github.com/aptly-dev/www.aptly.info and documenting this option ?
then we can merge it !

neolynx (approved):

nresare:
Thank you for merging! I am a little overwhelmed by work etc right now, but I have put on my todo list to propose some documentation changes, to be merged when this code exists in a released version

neolynx:
Thanks for the PR ! :-)

if you have some time to take care of the documentation, that would be awesome, let me know !

neolynx:
hey, just wanted to say thanks again, I like the feature a lot and new aptly will use it to publish itself ;-)

neolynx:
Btw: https://github.com/aptly-dev/aptly/discussions/1345

aptly is now using the -multi-dist to host itself :+1:

PR #1300, opened 2024-06-13
Enable SkipArchitectureCheck and IgnoreSignatures in mirror API

iofq:
Fixes #

The command line flags `-force-architectures` and `-ignore-signatures` are not consistently implemented in the mirror API like they are in the mirror CLI. Discovered because the example repo I'm attempting to use, `https://pkg.duosecurity.com/Ubuntu`, is non-standard and breaks without these specified.

## Requirements

## Description of the Change

`SkipArchitectureCheck` and `IgnoreSignatures` are implemented as params for the create/POST and update/PUT handlers under `/api/mirrors`. 

Did not implement unit tests as there doesn't appear to be a good harness for mocking downloads.

Updating functional tests would be nice, seems we can add to https://github.com/aptly-dev/aptly/blob/master/system/t12_api/mirrors.py if we potentially upload some system test files first. Would need some guidance to get started on that if desired. 

## Checklist

- [ ] documentation updated - See [Merge Request](https://github.com/aptly-dev/www.aptly.info/pull/109)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -58,3 +58,4 @@ List of contributors, in chronological order:
 * Ryan Gonzalez (https://github.com/refi64)
 * Paul Cacheux (https://github.com/paulcacheux)
 * Nic Waller (https://github.com/sf-nwaller)
+* iofq (https://github.com/iofq)
--- a/api/mirror.go
+++ b/api/mirror.go
@@ -52,19 +52,20 @@ func apiMirrorsList(c *gin.Context) {
 func apiMirrorsCreate(c *gin.Context) {
 	var err error
 	var b struct {
-		Name               string `binding:"required"`
-		ArchiveURL         string `binding:"required"`
-		Distribution       string
-		Filter             string
-		Components         []string
-		Architectures      []string
-		Keyrings           []string
-		DownloadSources    bool
-		DownloadUdebs      bool
-		DownloadInstaller  bool
-		FilterWithDeps     bool
-		SkipComponentCheck bool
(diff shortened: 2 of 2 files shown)


neolynx:
awesome work, thanks !

but I wonder, if apt can handle this repo, should aptly not also be able to mirror this weird structure ?

iofq:
> but I wonder, if apt can handle this repo, should aptly not also be able to mirror this weird structure ?

I think aptly is behaving as expected here, aptly by default checks that the components and archs defined for the mirror are present in the Release file, and this repo does not list any components or archs in its Release file. The skip component and arch check are available to override that check. 

[This document states:](https://wiki.debian.org/DebianRepository/Format) "If a server is not specifying the supported architectures with this field client behavior is unspecified in this case.". It does not mention anything about `Components` field requirements. So it seems there are no standards for this case, and what we're seeing is aptly default behavior is actually more strict than apt itself. 

Maybe aptly could be less strict about checking by default? Either move the checks from opt-out to opt-in, or perhaps check but only throw a warning, not an err.

neolynx:
I added a system test for SkipArchitectureCheck and IgnoreSignatures, see branch pr-1300

Maybe you can cherry-ick the commit or 'Allow edits from maintainers.' (https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork), so I could push it to the PR...

Then we are ready to  merge :)

PR #1304, opened 2024-06-17
Etcd Database Support

neolynx:
### etcd database support for aptly

replaces leveldb per configuration:
```
   "databaseBackend": {
       "type": "etcd",
        "url": "127.0.0.1:2379"
   }
```

Original PR: https://github.com/aptly-dev/aptly/pull/1023 (continued here due to no rebase permissions)

Contents:
- original PR
- fix int overflow
  - fixes goxz compile error
- fix db config and test
  -  fix unit test for adapted config file
- implement separate system tests
  - add t13_etcd test directory 
  - etcd will be started for the unit tests and each system test 
  - etcd will load fixture DB export if requested by the test
  - existing tests are reused for etcd testing
- implement temporary db support
  - temporary db support is implemented with a unique key prefix
  - prevent closing etcd connection when closing temporary db 
- implement batch operations
  - cache the operations internally in a list
  - Write() applies the list to etcd
- implement transactions
  - use temporary db for lookups in transactions
  - use batch implementation to commit transaction
- fixed expired ppa mirror key


neolynx:
The original implementation uses the main etcd db when when creating a TemporaryDB, which is then closed after usage and further db operation fail. 

Needs implementation for etcd temporary database, by using a prefix or so.

database/etcddb/database.go
```diff
@@ -0,0 +1,45 @@
...
+		MaxCallRecvMsgSize:   (2048 * 1024 * 1024) - 1,
+		DialKeepAliveTimeout: 7200 * time.Second,
+	}
+
+	cli, err := clientv3.New(cfg)
+	if err != nil {
+		return nil, err
+	}
+
+	return cli, nil
```
an0rak-dev, line 27: We could replace this with `return clientv3.New(cfg) since the library returns a `nil` client in case of error

database/etcddb/database.go
```diff
@@ -0,0 +1,45 @@
...
+
+func NewDB(url string) (database.Storage, error) {
+	cli, err := internalOpen(url)
+	if err != nil {
+		return nil, err
+	}
+	return &EtcDStorage{url, cli, ""}, nil
+}
+
+func NewOpenDB(url string) (database.Storage, error) {
```
an0rak-dev, line 38: What's the point of this function ? Since it can simply be resumed as 

```go
func NewOpenDB(url string) (database.Storage, error) {
   return NewDB(url)
}
```

Which adds no value to a direct call of `NewDB`

database/etcddb/storage.go
```diff
@@ -0,0 +1,205 @@
...
+func (s *EtcDStorage) applyPrefix(key []byte) []byte {
+	if len(s.tmpPrefix) != 0 {
+		return append([]byte(s.tmpPrefix+"/"), key...)
+	}
+	return key
+}
+
+// Get key value from etcd
+func (s *EtcDStorage) Get(key []byte) (value []byte, err error) {
+	key = s.applyPrefix(key)
```
an0rak-dev, line 36: Wouldn't be clearer to use another name here ? Like `prefixedKey` ?

database/etcddb/storage.go
```diff
@@ -0,0 +1,205 @@
...
+func (s *EtcDStorage) HasPrefix(prefix []byte) bool {
+	prefix = s.applyPrefix(prefix)
+	getResp, err := s.db.Get(Ctx, string(prefix), clientv3.WithPrefix())
+	if err != nil {
+		return false
+	}
+	if getResp.Count != 0 {
+		return true
+	}
+	return false
```
an0rak-dev, line 116: Could be replaced as `return getResp.Count > 0`

an0rak-dev (changes requested):

database/etcddb/database.go
neolynx: I think this stems from the goleveldb implementation, as it provides both too. Interestingly, the NewOpenDB is only used on go tests and should maybe be replaced...

context/context.go
```diff
@@ -288,7 +290,19 @@ func (context *AptlyContext) _database() (database.Storage, error) {
 	if context.database == nil {
 		var err error
 
-		context.database, err = goleveldb.NewDB(context.dbPath())
+		if context.config().DatabaseBackend.Type == "etcd" {
```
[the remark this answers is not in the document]
neolynx: done

context/context.go
```diff
@@ -288,7 +290,19 @@ func (context *AptlyContext) _database() (database.Storage, error) {
...
+			context.database, err = etcddb.NewDB(context.config().DatabaseBackend.URL)
+		} else if context.config().DatabaseBackend.Type == "leveldb" {
+			if context.config().DatabaseBackend.DbPath != "" {
+				dbPath := filepath.Join(context.config().RootDir, context.config().DatabaseBackend.DbPath)
+				context.database, err = goleveldb.NewDB(dbPath)
+			} else {
+				return nil, errors.New("leveldb databaseBackend config invalid")
+			}
```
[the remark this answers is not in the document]
neolynx: done

database/etcddb/database.go
```diff
@@ -0,0 +1,45 @@
...
+	clientv3 "go.etcd.io/etcd/client/v3"
+)
+
+var Ctx = context.TODO()
+
+func internalOpen(url string) (*clientv3.Client, error) {
+	cfg := clientv3.Config{
+		Endpoints:            []string{url},
+		DialTimeout:          30 * time.Second,
+		MaxCallSendMsgSize:   (2048 * 1024 * 1024) - 1,
```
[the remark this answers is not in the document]
neolynx: done

database/etcddb/database.go
neolynx: done

database/etcddb/storage.go
neolynx: maybe, however, the prefix is only added for temdbs. maybe applyPrefix should be maybePrefix ?

database/etcddb/storage.go
neolynx: done

database/etcddb/storage.go
an0rak-dev: Quite not a fan of `maybe` vars, it let us think that the behaviour might not be the one intended.

What do you think of `wantedKey` or `realKey` ?

database/etcddb/database.go
an0rak-dev: I leave the final decision to you 😄

neolynx: I removed the NewOpenDB from the etcddb module, leaving the goleveldb as it is

an0rak-dev (approved):

PR #1311, opened 2024-07-10
Fix bug in dependency resolution

### LINKED ISSUE #1312: Bug in dependency resolution with -filter-with-deps
There are cases when Aptly does not include all dependencies when `-filter-with-deps` is used.

## Detailed Description

Aptly has a bug in its dependency resolution where certain dependencies do not get added even if they are not satisfied.

Steps to reproduce (tested with latest `master`):
```shell
git clone ...
go build -o aptly.dev main.go
./aptly.dev mirror create -ignore-signatures -filter='python3-pep517|python3|black' -architectures=amd64 -filter-with-deps test1 https://snapshot.debian.org/archive/debian/20240707T211122Z/ bookworm
./aptly.dev mirror update -ignore-signatures test1
./aptly.dev -with-packages mirror show test1 | grep python3-tomli
```

**The package `python-pep517` depends on `python3-tomli`:**
```
Package: python3-pep517
Source: pep517
Version: 0.13.0-2
Installed-Size: 82
Maintainer: Debian Python Team <team+python@tracker.debian.org>
Architecture: all
Depends: python3-importlib-metadata, python3-tomli, python3-zipp, python3:any
[...]
```
**But Aptly does not include `python3-tomli` in the mirror.**

This is because `black` includes the following dependency: `python3-tomli | python3 (>> 3.11)`.
This dependency is satisfied by including `python3`, but as a side effect of evaluating this dependency, Aptly includes `python3-tomli` in the `cache` in `VerifyDependencies`.
The bug is that this causes `python3-tomli` to be excluded, even though its cache entry is `false` (i.e., not satisfied).

## Context

This is important for us because we rely on all dependencies being present when using `-filter-with-deps`.

## Possible Implementation

I proposed a fix in #1311.

## Your Environment

It's a logic error, so I do not think it varies based on distro, etc.
We could reproduce it on Debian bookworm as well as Arch Linux with the latest `master` branch of Aptly.

5hir0kur0:
Fixes #1312.

Aptly has a bug in its dependency resolution where certain dependencies do not get added even if they are not satisfied.

Steps to reproduce (tested with latest `master`):
```shell
git clone ...
go build -o aptly.dev main.go
./aptly.dev mirror create -ignore-signatures -filter='python3-pep517|python3|black' -architectures=amd64 -filter-with-deps test1 https://snapshot.debian.org/archive/debian/20240707T211122Z/ bookworm
./aptly.dev mirror update -ignore-signatures test1
./aptly.dev -with-packages mirror show test1 | grep python3-tomli
```

**The package `python-pep517` depends on `python3-tomli`:**
```
Package: python3-pep517
Source: pep517
Version: 0.13.0-2
Installed-Size: 82
Maintainer: Debian Python Team <team+python@tracker.debian.org>
Architecture: all
Depends: python3-importlib-metadata, python3-tomli, python3-zipp, python3:any
[...]
```
**But Aptly does not include `python3-tomli` in the mirror.**

This is because `black` includes the following dependency: `python3-tomli | python3 (>> 3.11)`.
This dependency is satisfied by including `python3`, but as a side effect of evaluating this dependency, Aptly includes `python3-tomli` in the `cache` in `VerifyDependencies`.
The bug is that this causes `python3-tomli` to be excluded, even though its cache entry is `false` (i.e., not satisfied).

## Requirements

## Description of the Change

This was causing issues for us because Aptly did not download all dependencies even with `-filter-with-deps`.

I fixed this bug by just checking for the `satisfied` status instead of also requiring that the dependency not be included in the cache.
```diff
diff --git a/deb/list.go b/deb/list.go
index ec14bb98..89fc7d08 100644
--- a/deb/list.go
+++ b/deb/list.go
@@ -351,7 +351,7 @@ func (l *PackageList) VerifyDependencies(options int, architectures []string, so
                                                cache[hash] = satisfied
                                        }

-                                   if !satisfied && !ok {
+                                 if !satisfied {
                                                variantsMissing = append(variantsMissing, dep)
                                        }

```

## Checklist

- [ ] documentation updated


neolynx:
thanks a lot for catching and fixing this !

looks like the tests were relying on duplicated packages, hence they fail now...

neolynx:
I rebased and updated the tests. 

Could you maybe check whether these changes are OK ? the original code seems to have missed dependencies on bind, emacs22, kde, ttf-thryomanes, xpdf-reader. Since some of them are meta packages, is this really the desired dependency filtering ?

5hir0kur0:
@neolynx 

Thank you! Looks good to me.

As for the question about virtual packages, I am not sure, maybe it depends on how the callers of `VerifyDependencies` handle virtual packages?

But in general I think that it has been possible for virtual/meta packages to be returned from `VerifyDependencies` since before my change, so this should not be a new phenomenon.

neolynx (approved):

PR #1316, opened 2024-07-16
fix api tests and output

neolynx:
## Description of the Change

The ReposAPITestCopyPackage sporadically fails (i.e `\u003cEOL\u003e` (<EOL>) as filename).

This change fixes the implementation of apiReposCopyPackage, making sure the query parameter is read outside of `maybeRunTaskInBack`.

Also, check if there are packages to copy in ReposAPITestCopyPackage or fail with an error. 

This change also improves the error output in case a test fails, and enables s3 tests locally when running `make docker-system-tests`. Also, some lost tests were reenabled and others fixed.


Makefile
```diff
@@ -58,7 +58,10 @@ endif
 
 docker-test: install
 	go test -v -coverpkg="./..." -c -tags testruncli
-	PATH=$(BINPATH)/:$(PATH) APTLY_VERSION=$(VERSION) $(PYTHON) system/run.py --long $(TESTS) --coverage-dir $(COVERAGE_DIR) $(CAPTURE) $(TEST)
+	test -e aws.creds && . ./aws.creds; \
```
an0rak-dev, line 61: Is this file automatically created during the `make docker-build-system-tests` (or any other dependant) target ?

system/t12_api/snapshots.py
```diff
@@ -208,7 +213,7 @@ def check(self):
                  "Signing": DefaultSigningOptions,
             }
         )
-        self.check_equal(resp.json()['State'], TASK_SUCCEEDED)
+        self.check_task(task)
 
         self.check_equal(self.delete_task("/api/snapshots/" + snap2).json()['State'], TASK_FAILED)
         self.check_equal(self.delete_task("/api/snapshots/" + snap2, params={"force": "1"}).json()['State'], TASK_FAILED)
```
an0rak-dev, line 219: Why not replacing occurences of `check_equal(..., TASK_FAILED)` with `check_task_fail` in this file ?

an0rak-dev (changes requested):

Makefile
neolynx: no, the file has to be place there manually, if it is there s3 tests will be enabled..

system/t12_api/snapshots.py
neolynx: fixed, thanks !  :)

Makefile
an0rak-dev: Should we not indicate it in the ReadMe or in the project's wiki ?

neolynx: tthis is true, alongside with openstack swift and azure testing...

I will move it to a next PR, has nothing to do in here.

an0rak-dev: Sounds great.

Let me know when you have moved the aws.creds part, and this PR will be Ok for me 👌🏻

neolynx: I moved it our of this PR

an0rak-dev (approved):

PR #1317, opened 2024-07-18
Support version numbers in `Provides` entries

### LINKED ISSUE #678: Packages with "Provides" tags
Today, I tried to add "python-cryptography" (from ubuntu bionic) to my aptly mirror filter (configured to pull dependencies).
When I try to install this package, apt claims two dependencies as not available:
python-cffi-backend-api-min
python-cffi-backend-api-max

    $ apt-cache show python-cryptography
    Package: python-cryptography
    [...]
    Depends: [...], python-cffi-backend-api-min (<= 9729), python-cffi-backend-api-max (>= 9729), [...]
    [...]

I didn't find those packages even on official ubuntu repository. After some seeking, I recognized those packages are provided by "python-cffi-backend".

    $ apt-cache show python-cffi-backend 
    Package: python-cffi-backend
    [...]
    Provides: python-cffi-backend-api-9729, python-cffi-backend-api-max (= 10239), python-cffi-backend-api-min (= 9729)
    [...]

Is it possible to take "Provides" tags into dependency resolution?

5hir0kur0:
Fixes #678, #1318

## Requirements

## Description of the Change

Aptly was interpreting the entries of the `Provides` field as plain package names, but they are allowed to contain version numbers.

> The relations allowed are [...]. The exception is the Provides field, for which only `=` is allowed.
> [...]
> A Provides field may contain version numbers, and such a version number will be considered when
> considering a dependency on or conflict with the virtual package name.
– https://www.debian.org/doc/debian-policy/ch-relationships.html

We ran into the same issue as #678, funnily enough with the exact same package.
Here's a quick way to reproduce the issue (tested with the current `master`):
```shell
go build -o aptly.dev main.go
./aptly.dev mirror drop test1
./aptly.dev mirror create -ignore-signatures -architectures=amd64 -filter-with-deps -filter="python3-cryptography" test1 'https://snapshot.debian.org/archive/debian/20240717T025830Z/' bookworm
./aptly.dev mirror update -ignore-signatures -dep-verbose-resolve test1
./aptly.dev mirror show -with-packages test1 | grep python3-cffi-backend
```
The package `python3-cffi-backend` should be included in the mirror, but on the `master` branch it is not and the following warning is printed:
```
Unsatisfied dependency: python3-cffi-backend-api-min (<= 9729) [amd64]
Unsatisfied dependency: python3-cffi-backend-api-max (>= 9729) [amd64]
Missing dependencies: python3-cffi-backend-api-min (<= 9729) [amd64], python3-cffi-backend-api-max (>= 9729) [amd64]
```

I fixed this by parsing the `Provides` entries in the places where they were used as plain strings.

For reference, this is what the `Provides` entries of `python3-cffi-backend` look like:
```
Package: python3-cffi-backend
[...]
Provides: python3-cffi-backend-api-9729, python3-cffi-backend-api-max (= 10495), python3-cffi-backend-api-min (= 9729)
```

## Checklist

- [ ] documentation updated


5hir0kur0:
So, my theory for why some tests are failing is that it's because I changed the `break`s here to `return`s:
```go
// Search searches package index for specified package(s) using optimized queries
func (l *PackageList) Search(dep Dependency, allMatches bool) (searchResults []*Package) {
	if !l.indexed {
		panic("list not indexed, can't search")
	}

	i := sort.Search(len(l.packagesIndex), func(j int) bool { return l.packagesIndex[j].Name >= dep.Pkg })

	for i < len(l.packagesIndex) && l.packagesIndex[i].Name == dep.Pkg {
		p := l.packagesIndex[i]
		if p.MatchesDependency(dep) {
			searchResults = append(searchResults, p)

			if !allMatches {
				break // <-- here
			}
		}

		i++
	}

	if dep.Relation == VersionDontCare {
		for _, p := range l.providesIndex[dep.Pkg] {
			if dep.Architecture == "" || p.MatchesArchitecture(dep.Architecture) {
				searchResults = append(searchResults, p)

				if !allMatches {
					break // <-- and here
				}
			}
		}
	}

	return
}
```
([deb/list.go](https://github.com/aptly-dev/aptly/blob/8029305d321fca430dd84d986ea4c3ca1443b0d9/deb/list.go#L462))

I did this because there are two `for`-loops here that search for matches and if `allMatches` is `false`, from what I understand it should be enough for one match to be returned.
However, the way it's currently implemented it would return two matches if both `for` loops find a match.

However, this is kind of unrelated to the `Provides` issue (I just had to change the `Search` function a bit anyway to support versions in `Provides`...).
If replacing the `break`s with `return`s is wrong I'll change it back to the old behavior.

deb/package.go
```diff
@@ -321,24 +338,56 @@ func (p *Package) MatchesArchitecture(arch string) bool {
...
 	if dep.Architecture != "" && !p.MatchesArchitecture(dep.Architecture) {
 		return false
 	}
 
-	if dep.Relation == VersionDontCare {
-		if utils.StrSliceHasItem(p.Provides, dep.Pkg) {
-			return true
-		}
-		return dep.Pkg == p.Name
+	if providesDep, _ := p.providesDependency(dep); providesDep { // Is ignoring errors the right thing to do here?
```
neolynx, line 378: should probably be logged as warning...

neolynx:
Thanks a lot for your work !

Could you please rebase on master and maybe give me permission to rebase your branch ?

I added some changes, see https://github.com/aptly-dev/aptly/pull/1325:
- it makes sense to collect all the errors in `providesDependency`, instead of abortind the dependency search on first error. 
- I made it go 1.19 compatible

it looks food so far, but it removes all pure virtual packages from the tests (i.e. output of aptly snapshot verify). We need to check if this is the desired behavior...

neolynx:
> I did this because there are two `for`-loops here that search for matches and if `allMatches` is `false`, from what I understand, it should be enough if just _one_ match is returned. However, the way it's currently implemented it would return _two_ matches if both `for` loops find a match.

that might be the reason for not getting the full virtual packages...

5hir0kur0:
> Could you please rebase on master and maybe give me permission to rebase your branch ?

OK, I'll rebase it. I gave you access to my fork now as well.
It's weird that checking this box apparently was not enough to allow you to rebase the branch (?)
image

deb/package.go
5hir0kur0: OK, I added a log message there.

I wasn't sure if I should use `"github.com/rs/zerolog/log"` or `"log"` (from the standard library) as both seem to be used in Aptly. I went with the former.

neolynx:
could you try running the test in docker > unfortunately a lot of stuff depends on the environment/path and so on to be setup correctly...

see https://github.com/aptly-dev/aptly/blob/master/CONTRIBUTING.md#docker-development-setup (new content)

5hir0kur0:
Thanks, with `make docker-system-tests` I am seeing the same results as the CI pipeline.
"t05_snapshot" and "t09_repo" fail with my changes.
## t05_snapshot

Here there are a bunch of packages that are no longer in `Missing dependencies` in the output of `snapshot verify`. All the missing packages seem to be packages that are provided by other packages with a `Provides:` entry. I think they are just missing from the "missing dependencies" because they are actually provided by other packages and thus not a missing dependency.
## t09_repo: ImportRepo1Test

```
running command: aptly repo import wheezy-main repo1 'nginx (>= 1.2)' unpaper_0.4.2-1_amd64

Traceback (most recent call last):
  [...]
Exception: content doesn't match:
---
+++
@@ -1,3 +1,12 @@
 Loading packages...
+[o] nginx-extras_1.2.1-2.2+wheezy2_amd64 imported
+[o] nginx-extras_1.2.1-2.2+wheezy2_i386 imported
+[o] nginx-full_1.2.1-2.2+wheezy2_amd64 imported
+[o] nginx-full_1.2.1-2.2+wheezy2_i386 imported
+[o] nginx-light_1.2.1-2.2+wheezy2_amd64 imported
+[o] nginx-light_1.2.1-2.2+wheezy2_i386 imported
+[o] nginx-naxsi-ui_1.2.1-2.2+wheezy2_all imported
+[o] nginx-naxsi_1.2.1-2.2+wheezy2_amd64 imported
+[o] nginx-naxsi_1.2.1-2.2+wheezy2_i386 imported
 [o] nginx_1.2.1-2.2+wheezy2_all imported
 [o] unpaper_0.4.2-1_amd64 imported
```

It seems like the command `aptly repo import wheezy-main repo1 'nginx (>= 1.2)'` now imports more packages than before. The way I understand it, this command is saying "import all packages matching the query `nginx (>= 1.2)` from `wheezy-main` into `repo1`" (correct me if I'm wrong). Since all of the additional packages provide `nginx`, I would consider the new behavior to be correct:
```shell
$ grep -EA15 '^Package: (nginx-extras|nginx-full|nginx-light|nginx-naxsi)$' Packages | grep Provides
Provides: httpd, nginx
Provides: httpd, nginx
Provides: httpd, nginx
Provides: httpd, nginx
```


What do you think @neolynx?

neolynx:
Thanks for the analysis, [5hir0kur0](https://github.com/5hir0kur0) !

For `t09_repo: ImportRepo1Test` I would agree, this is likely the correct behavior. I assume if one would not want those to be mirrored/imported, a filter rule could exclude them. 

For `t05_snapshot` can we somehow prove that the providing packages are there ? I would like to be sure why they do not show up in the missing dependencies. I poekd around a bit, by adding `sleep 3000` to the fixutre command and inspect the repo, but I was not getting far, since for the test aptly seems to have only the db, but not the pool data...

neolynx (approved):

neolynx:
@5hir0kur0 I do not have your email address, would you mind dropping me a mail to my commit address ?

PR #1340, opened 2024-09-18
Update CI release builds to remove goxc

iofq:
Fixes #

## Requirements

## Checklist

- [ ] documentation updated

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,13 +94,14 @@ jobs:
           files: coverage.txt
 
   ci-debian-build:
-    name: "CI Build"
+    name: "Build"
     needs: test
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
         name: ["Debian 13/testing", "Debian 12/bookworm", "Debian 11/bullseye", "Debian 10/buster", "Ubuntu 24.04", "Ubuntu 22.04", "Ubuntu 20.04"]
+        arch: ["amd64", "i386" , "arm64" , "armhf"]
         include:
           - name: "Debian 13/testing"
             suite: trixie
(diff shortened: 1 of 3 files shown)


neolynx:
I am not a big fan of doing the lint in the same pipeline, as the system tests run for 20 minutes and they can run even if lint complains. I moved the lint to a separate check, so it can run in parallel and same me a lot of time... :-)

.github/workflows/ci.yml
```diff
@@ -212,9 +222,50 @@ jobs:
...
-      - name: "Build goxc packages"
+      - name: "Build packages for ${{ matrix.goos }} ${{ matrix.goarch }}"
         env:
           GOBIN: /usr/local/bin
         run: |
+          set -ex
+          mkdir -p tmp/man tmp/completion/bash_completion.d tmp/completion/zsh/vendor-completions
           go version
-          make release
+          echo ${{ github.ref_name }} > VERSION
```
neolynx, line 232: we should use `make version` here to have the same version

.github/workflows/ci.yml
```diff
@@ -53,7 +48,18 @@ jobs:
 
       - name: "Get aptly version"
         run: |
-           make version
+           make version > VERSION
```
neolynx, line 51: the `VERSION` file is generated in debina/rules and the Makefile, the pipeline should avoid writing to the git repo directly I think...

neolynx:
i pulled in some of your changes to test it ...  thanks a lot !

iofq:
> https://github.com/aptly-dev/aptly/actions/runs/10919177530/job/30307362665
> 
> seems it cannot find the $path somehow

hmm seems like it's an issue with escaping slashes in the branch/ref name, we may have to replace slashes with dashes in ref_name

neolynx:
I got it somehow to build on my branch (you might wanna rebase). however the contects of the zip files seem wrong, if you wann ahave a look at it...

for the lint part, did I assume right that the whole pipeline (i.e. system tests) will not run if link fails ?

iofq:
> for the lint part, did I assume right that the whole pipeline (i.e. system tests) will not run if link fails ?

Yeah that was my intention, if linting is going to fail anyway then we should skip the 20 minute unit tests, but if you feel otherwise that's fine with me

neolynx:
I would like to get both lint and system tests results for PRs directly, otherwise I might have to fix lint errors before seeing if it actually works, as it happened before. 

Merging will not be done with a failed lint run anyway...

neolynx:
looks great ! I cherry-picked and updated slightly, the versions are now consistently based on the debian version, and release versions are used when a tag is present. 

 aptly_1.6.0~beta1+20240919091930.86e0512e_freebsd_arm.zip.zip 

the artifacts seem to be double zipped still, and we need to check if the contents are exactly the same as before...

neolynx:
The contents of the original archived look as follows:

```
./aptly_1.5.0_linux_amd64:
total 28568
-rwxr-xr-x 1 lynx lynx 29231726 Jun 28  2022 aptly
-rw-r--r-- 1 lynx lynx     2268 Jun 28  2022 AUTHORS
-rw-r--r-- 1 lynx lynx     1092 Jun 28  2022 LICENSE
drwxr-xr-x 2 lynx lynx     4096 Sep 19 13:10 man
-rw-r--r-- 1 lynx lynx     4340 Jun 28  2022 README.rst

./aptly_1.5.0_linux_amd64/man:
total 52
-rw-r--r-- 1 lynx lynx 52190 Jun 28  2022 aptly.1
```

this is from `aptly_1.5.0_linux_amd64.tar.gz`. the contents of1`aptly_1.5.0_darwin_amd64.zip` and `aptly_1.5.0_freebsd_amd64.zip` look the same.

iofq:
> the artifacts seem to be double zipped still, and we need to check if the contents are exactly the same as before...

The double-zip is related to https://github.com/actions/upload-artifact/issues/39 and doesn't end up in the final release zips like [these](https://github.com/iofq/aptly/releases/tag/v1.35.0) Based on the issue there isn't a good workaround, maybe could try a different upload/download action?


> The contents of the original archives look as follows:

I am happy to mirror the layout of the previous release archives; do you have interest in adding the completion files to the archives moving forward?

neolynx:
Seems to work now on my branch: https://github.com/aptly-dev/aptly/actions/runs/10944845975/job/30388887371

I got rid of the double zip by unzipping the go build output, and I added the missing files. I think the architves lok ok now.

Also the matrix build is now properly used for the debian packages, following your example. Also updated the version handling a bit, to prepare testing a release verion without CI suffix...

iofq:
> I got rid of the double zip by unzipping the go build output, and I added the missing files. I think the architves lok ok now.

It looks like the release doesn't include any files now. I think you can't just upload folders to the release, you need archives (the double zip doesn't happen on the actual release page)

PR #1352, opened 2024-10-02
Feature/storage api

neolynx:
## Description of the Change

- **add storage API - Get disk free info of aptly storage**
- `make docker-deb`: install and init swagger
- improve `make docker-shell`
- system-tests: install swag only if needed
- make swagger quiet
- fix a disabled system test depending on gpg1 keys
- fix a disabled system test not really depending onFTP
- go.etcd.io/etcd modules updated to match debian version
- improve ci output


cfiehe (approved):
Looks good to me. Let's give it a try 😃.

Issue #1338, opened 2024-09-13
Update Publish does not suport slash in prefix


neolynx:
## Detailed Description

When updating a publish (via api PUT /publish/:prefix/:distribution ) the prefix cannot contain a slash.

Probably applies to other operations (create publish, ...) as well.

## Context

On the command line, aprly can create a publish point containing a slash, but api fails to operate on it. 

## Possible Implementation

Parse the arguments in golang, instead of gin router / use json data for distribution

## Your Environment

latest CI.

PR #1356, opened 2024-10-02
Allow to add/switch components of a published local repository

### LINKED ISSUE #1355: Allow to add/update components of a published local repository
## Detailed Description
Besides the capability to add/update components of a published snapshot-based repository, it would be useful if adding/updating components of published local repositories would be possible, too.

## Context
Adding and updating components of a published local repository is currently only possible by dropping and recreating the whole published repository. This causes unnecessary overhead, which can be avoided if components and their package sources become fully manageable for published local repositories.

## Possible Implementation
Modification of the `switch` command in the CLI, so that adding/updating components of published local repositories can also be done. Update of the REST API by introducing a new generalized `Sources` attribute for the PUT command on a published repository, in order to add/update components of a published local repository.

cfiehe:
Fixes #1355, https://github.com/aptly-dev/aptly/issues/1338

## Description of the Change

This commit modifies the behavior of the publish `switch` method in the way, that it can also be used to add/update components of published local repositories. Furthermore, the commit ensures that the sources returned by a published repository are ordered by their component names.

## Checklist

- [x] documentation updated
- [x] make sure slash can be used in prefix (https://github.com/aptly-dev/aptly/issues/1338)
- [ ] swagger documentation


neolynx:
maybe changing the logging should not be done if not really neccessary, since the aptly output is used extensively in the testing.

if you wanna update the tests with the new logs, uncomment CAPTURE in the Makefile ;-)

cfiehe:
Thanks a lot. I was not aware of the consequences. Unfortunately, we must change the output of the switch command and make it more generic, so that it can be also used for published local repositories. It requires some time for fixing the test cases.

neolynx:
amazing work :)

I sent you an invitation as maintainer, it would be great to have more of your contribution as well as reviewing some PRs (i.e. https://github.com/aptly-dev/aptly/pull/1352) to move things forward a bit faster !

neolynx:
I dont have the rights to push to your branch, but here would be the golang-lint fix: https://github.com/aptly-dev/aptly/commit/eb15a656bb5c76b57b76caf695229b8de72b675a :-)

neolynx:
there is one more log update in a S3 test (where your i does not have creds): https://github.com/aptly-dev/aptly/actions/runs/11153050817/job/30999844378#step:9:1456

cfiehe:
@neolynx
Have you got an idea, why the test execution is failing?
```
--- PASS: Test (0.01s)
PASS
coverage: 74.9% of statements
ok  	github.com/aptly-dev/aptly/utils	0.011s	coverage: 74.9% of statements
FAIL

Stopping etcd ...

Unit Tests FAILED
```

cfiehe:
> I dont have the rights to push to your branch, but here would be the golang-lint fix: [eb15a65](https://github.com/aptly-dev/aptly/commit/eb15a656bb5c76b57b76caf695229b8de72b675a) :-)

Ok, that is strange. The flag that the maintainer has editing rights is set.

neolynx:
76.85% of diff hit (target 74.90%) 

nice !

cfiehe:
It finally turns into green 😅. Thanks a lot for your support.

api/publish.go
```diff
@@ -245,11 +289,22 @@ func apiPublishRepoOrSnapshot(c *gin.Context) {
...
+// @Failure 404 {object} Error "Published repository or source not found"
+// @Failure 500 {object} Error "Internal Error"
+// @Router /api/publish/{prefix}/{distribution} [put]
 func apiPublishUpdateSwitch(c *gin.Context) {
 	param := parseEscapedPath(c.Params.ByName("prefix"))
 	storage, prefix := deb.ParsePrefix(param)
-	distribution := c.Params.ByName("distribution")
+	distribution := parseEscapedPath(c.Params.ByName("distribution"))
 
 	var b struct {
```
neolynx, line 309: need to document the body json parameters: See https://github.com/aptly-dev/aptly/blob/feature/176-snapshot-pull-api/api/snapshot.go#L506

api/publish.go
```diff
@@ -363,14 +452,27 @@ func apiPublishUpdateSwitch(c *gin.Context) {
 	})
 }
 
-// DELETE /publish/:prefix/:distribution
+// @Summary Delete published repository
+// @Description Delete a published repository.
```
neolynx, line 456: make bold: **Delete a published repository.**

copy api and field description from here: https://www.aptly.info/doc/api/publish/

api/publish.go
```diff
@@ -83,7 +83,50 @@ func apiPublishList(c *gin.Context) {
...
+	if err != nil {
+		AbortWithJSONError(c, 500, fmt.Errorf("unable to show: %s", err))
+		return
+	}
+
+	c.JSON(200, published)
+}
+
+// @Summary Create published repository
+// @Description Create a published repository with specified parameters.
```
neolynx, line 120: document json params

neolynx:
Another more high level question: if we add a /remove in the other PR, shouldn't we add a /add here instead of modifying the existing update/switch ? 

We could also keep the switch/update not touching the snapshots/components, but rather remember the multi-dist, and add 2 new APIs for adding/removing them, maybe call them /extend and /redude ?

just thinking loudly...

neolynx:
maybe we can also make sure that a slash in prefix and distribution is handled correctly everywere ?

api/publish.go
```diff
@@ -245,11 +289,22 @@ func apiPublishRepoOrSnapshot(c *gin.Context) {
...
+// @Success 200 {object} deb.RemoteRepo
+// @Failure 400 {object} Error "Bad Request"
+// @Failure 404 {object} Error "Published repository or source not found"
+// @Failure 500 {object} Error "Internal Error"
+// @Router /api/publish/{prefix}/{distribution} [put]
 func apiPublishUpdateSwitch(c *gin.Context) {
 	param := parseEscapedPath(c.Params.ByName("prefix"))
 	storage, prefix := deb.ParsePrefix(param)
-	distribution := c.Params.ByName("distribution")
+	distribution := parseEscapedPath(c.Params.ByName("distribution"))
```
neolynx, line 307: great !

cfiehe:
> Another more high level question: if we add a /remove in the other PR, shouldn't we add a /add here instead of modifying the existing update/switch ?

Yes, I had the same thoughts. It "feels" a little bit asymmetric, having a `remove` command in the CLI and in the REST API, but no corresponding `add` command.
 
> We could also keep the switch/update not touching the snapshots/components, but rather remember the multi-dist, and add 2 new APIs for adding/removing them, maybe call them /extend and /redude ?

I am not sure, if I get that right. Before the enhancements, the `switch/update` in the REST API was used for switching snapshots of a published snapshot repository. So even before the change, this command touched snapshots/components, but it was limited to published snapshot repositories and could not be used for published local repositories.

We can introduce two new commands in the CLI publish group, e.g. `extend` and `reduce` (typo above?) or `add` and `remove` what you like more. These commands can be used to add new components to a published repository or remove existing components from a published repository. But how to handle updates? Should we stick to the existing `switch` command for changing the sources (snapshot, local repo) for components that are already part of a published repository?

In the REST API, we can introduce  `extend` and `reduce` or `add` and `remove` as new subresources of a published repository. But how to handle updates? The current approach is to use `PUT` on the managed resource, do you want to change that?

api/publish.go
```diff
@@ -52,10 +52,10 @@ func parseEscapedPath(path string) string {
 	return result
 }
 
-// @Summary Get publish points
-// @Description Get list of available publish points. Each publish point is returned as in “show” API.
+// @Summary Get published repositories
+// @Description Get a list of published repositories. Each published repository is returned as in "show" API.
```
neolynx, line 56: make bold with `**`

api/publish.go
neolynx: make description bold, not dot

api/publish.go
```diff
@@ -245,11 +289,22 @@ func apiPublishRepoOrSnapshot(c *gin.Context) {
 	})
 }
 
-// PUT /publish/:prefix/:distribution
+// @Summary Update published repository
```
neolynx, line 292: make bold

neolynx: copy multi line description from https://www.aptly.info/doc/api/publish/,

neolynx:
> I am not sure, if I get that right. Before the enhancements, the `switch/update` in the REST API and `switch` in the CLI were used for switching snapshots of a published snapshot repository. So even before the change, these commands touched snapshots/components, but they were limited to published snapshot repositories and could not be used for published local repositories.

ok, so the change here is that we also allow local repos. adding new snapshots was already possible in that since, right ?

>  `add` and `remove` what you like more

I think add/remove is better.

> But how to handle updates? Should we stick to the existing `switch` command for changing the sources (snapshot, local repo) for components that are already part of a published repository?

what I fail to wrap my head around: with aptly commands we have publish update and publish switch. In the API both are done with the PUT ? What is the difference between switch and update ?

> In the REST API, we can introduce `extend` and `reduce` or `add` and `remove` as new subresources of a published repository. But how to handle updates? The current approach is to use `PUT` on the published repository resource. Do you want to change that?

I would not change existing behavior, therefore just adding /remove is the least invasive. We can still add /add later.

cfiehe:
>ok, so the change here is that we also allow local repos. adding new snapshots was already possible in that since, right ?

Before we started, you had to define all components you want to have in your published repository at creation time. Afterwards, it was not possible to add or remove components from a published repository of any kind (snapshot or local). Adding or removing requires dropping the published repository and recreating it by which the new component set. That means the component set of a published repository was immutable and could not be modified after creation. The `switch` command allows you – only in case of a published snapshot repository – to replace a snapshot associated with a component. The `update` command tells the published repository – only in the case of a published local repository – to refresh the packages from the backed local repository. But it was not possible to associate an existing component with another local repository. In contrast to a published snapshot repository, the mapping between component and local repository was immutable and could not be changed after creation.

For example, we have defined a mirror of all packages in the `main` component of Ubuntu 24.04. Now, we create the snapshot `ubuntu-noble-main-1` and publish it in a published snapshot repository by associating the `main` component with the snapshot `main -> ubuntu-noble-main#1`. The switch command allows you to associate another snapshot with the `main` component, e.g.  `main -> ubuntu-noble-main#2`. Now you reach the point where you e.g. also want to have all packages in your published repository, which come from the `universe` component. You define a new mirror and create the snapshot `ubuntu-noble-universe#1`. The problem is, that the `switch` command can only operate on components that are already inside your published repository, you cannot use it to add the new `universe` component. The former approach was to drop your published repository and recreate with the sources `main -> ubuntu-noble-main#2` and `universe -> ubuntu-noble-universe#1`. After some time, you want to get rid of the `universe` component again, because the packages may be no longer required. You can use the `switch` command to associate e.g. an empty snapshot with the `universe` component, but you cannot actually drop it from your published repository without recreating the published repository.

What we are currently trying to achieve is, that Aptly offers a way to modify the component set after creation time by adding and removing components and by associating them with snapshots or local repositories, respectively. The following use case were not supported before our changes:
- Add components to a published repository of any kind
- Remove components from a published snapshot of any kind
- Associate a component with another local repository in case of a published local repository (the `switch` command already allowed that in case of published snapshot repository)

cfiehe:
> what I fail to wrap my head around: with aptly commands we have publish update and publish switch. In the API both are done with the PUT ? What is the difference between switch and update ?

A snapshot is an immutable package set. Calling `update` on a published snapshot repository will never pull in new packages. If you want to change your package set, you must `switch` to another snapshot which may contain a different package set. 

However, in case of a local repository someone could have added or removed packages. In order to make them accessible from the outside, you must advice the corresponding published local repository to re-fetch the package set from the backed local repositories.

cfiehe:
Summary:
- Introduce `publish add` to add new components to a publish repository
- Introduce `publish remove` to remove existing components from a publish repository
- Use `publish switch` to change the mapping between a component and its package source (snapshot/local repository)

cfiehe:
That would be an alternative REST-API but it is not so RESTful, because we actually model actions and it has the problem, that `switch` is currently done by via `PUT` on /api/publish/{prefix}/{distribution}`.
```
/api/publish/{prefix}/{distribution}/add[post]
/api/publish/{prefix}/{distribution}/switch[post]
/api/publish/{prefix}/{distribution}/remove[post]
```

cfiehe:
I have implemented the new `publish add` command and used the same CLI style, which is used for `publish switch`. `add` succeeds only when the component does not already exist in the published repository. I have modified the `publish switch` command, so that you can also switch the local repositories of a published local repository. The manual might need an extension.

I have simplified the REST API and introduced the new `Sources` attribute. Now a simple `PUT` with the entire source list triggers the update of the published repository. The only drawback is that in most cases the client needs to query the resource first in order to get the full source list. In the next step, the list is modified, so that it represents the entire target state. Sources, which are not in the list, but exist in the published repository, are removed. I think this approach is more RESTful and fits best in Aptly's REST API.

In the next day, I will have a look at the (Swagger) documentation and add what is missing.

deb/publish.go
```diff
@@ -452,8 +456,26 @@ func (p *PublishedRepo) UpdateLocalRepo(component string) {
 	p.rePublishing = true
 }
 
-// UpdateSnapshot switches snapshot for component
-func (p *PublishedRepo) UpdateSnapshot(component string, snapshot *Snapshot) {
+// UpsertLocalRepo inserts/updates local repository source for component
```
neolynx, line 459: typo

neolynx:
shall we clos this in favor of https://github.com/aptly-dev/aptly/pull/1366 ?

neolynx:
closing in favor of https://github.com/aptly-dev/aptly/pull/1366

PR #1364, opened 2024-10-07
Feature/persist multidist

neolynx:
## Description of the Change

The recently introduced `MultiDist` feature for publish points is now persisted to the database.

This allows specifying the `MultiDist` option when creating a publish point (`publish snapshot` / `publish repo`) and making it optional for `publish update` / `publish switch` (it may be provided to enable/disable the feature later on).


neolynx:
@cfiehe as discussed, maybe you wanna rebase your branch :)

api/publish.go
```diff
@@ -339,10 +339,14 @@ func apiPublishUpdateSwitch(c *gin.Context) {
 		published.AcquireByHash = *b.AcquireByHash
 	}
 
+	if b.MultiDist != nil {
+		published.MultiDist = *b.MultiDist
+	}
```
cfiehe, line 344: Should we not drop support for setting `MultiDist` after published repository creation at all?

cfiehe (changes requested):
I think, we should make `MultiDist` immutable after published repository creation.

api/publish.go
neolynx: I thought about it, but for existing publish points this will automatically also persist the setting, so we can depreciate it later (or keep it changeable)...

cfiehe: Ah, ok. Let's do the persistence first and discuss immutability later 😃.

cfiehe (approved):

PR #1366, opened 2024-10-08
Allow adding, removing and replacing of published repository components

cfiehe:
Fixes: https://github.com/aptly-dev/aptly/issues/1338, https://github.com/aptly-dev/aptly/issues/1349, https://github.com/aptly-dev/aptly/issues/1355

## Description of the Change

This commit allows to add, remove and update components of published repositories without the need to recreate them.

## Checklist

- [x] documentation updated


cfiehe:
Green is the best color 😄.

system/t06_publish/PublishUpdate10Test_gold
```diff
@@ -1,10 +1,8 @@
-WARNING: force overwrite mode enabled, aptly might corrupt other published repositories sharing the same package pool.
```
neolynx, line 1: why is the warning gone ?

system/t06_publish/PublishUpdate14Test_gold
```diff
@@ -1,8 +0,0 @@
-Loading packages...
```
neolynx, line 1: why is this test gone ?

system/t06_publish/PublishUpdate6Test_gold
```diff
@@ -1 +1,8 @@
-ERROR: unable to update: not a local repository publish
```
neolynx, line 1: this test was supposed to fail, no ?

system/t06_publish/PublishUpdate7Test_gold
```diff
@@ -1,8 +1,6 @@
 Loading packages...
 Generating metadata files and linking package files...
 Finalizing metadata files...
-Signing file 'Release' with gpg, please enter your passphrase when prompted:
```
neolynx, line 4: why is signing gone ?

system/t06_publish/PublishUpdate8Test_gold
```diff
@@ -1,6 +1,3 @@
 Loading packages...
 Generating metadata files and linking package files...
-Finalizing metadata files...
-Cleaning up prefix "." components contrib, main...
-
-Publish for local repo ./squeeze [i386] publishes {contrib: [repo2]}, {main: [repo1]} has been successfully updated.
+ERROR: unable to publish: unable to process packages: error linking file to ${HOME}/.aptly/public/pool/main/p/pyspi/pyspi_0.6.1.orig.tar.gz: file already exists and is different
```
neolynx, line 3: caution, this test did not fail before

system/t06_publish/PublishUpdate9Test_gold
```diff
@@ -1,3 +1,10 @@
+WARNING: force overwrite mode enabled, aptly might corrupt other published repositories sharing the same package pool.
+
 Loading packages...
 Generating metadata files and linking package files...
-ERROR: unable to publish: unable to process packages: error linking file to ${HOME}/.aptly/public/pool/main/p/pyspi/pyspi_0.6.1.orig.tar.gz: file already exists and is different
```
neolynx, line 3: test is supposed to fail...

system/t06_publish/update.py
```diff
@@ -217,20 +217,6 @@ class PublishUpdate5Test(BaseTest):
...
-    fixturePool = True
-    fixtureCmds = [
-        "aptly snapshot create snap1 from mirror gnuplot-maverick",
-        "aptly publish snapshot -keyring=${files}/aptly.pub -secret-keyring=${files}/aptly.sec snap1",
-    ]
-    runCmd = "aptly publish update maverick"
-    expectedCode = 1
-
-
-class PublishUpdate7Test(BaseTest):
```
neolynx, line 233: where did test 7 go ?

cfiehe: Everything is ok. Unfortunately, the following test case does not make sense any longer, because the update now also works for published snapshot repositories:
```
class PublishSwitch6Test(BaseTest):
    """
    publish switch: not a snapshot
    """
    fixtureCmds = [
        "aptly snapshot create snap1 empty",
        "aptly repo create local-repo",
        "aptly repo add local-repo ${files}",
        "aptly publish repo -keyring=${files}/aptly.pub -secret-keyring=${files}/aptly.sec -distribution=maverick local-repo",
    ]
    runCmd = "aptly publish switch maverick snap1"
    expectedCode = 1
```
I dropped that test case and renamed the subsequent cases.

cfiehe: We can add additional test cases at the end for the new features.

system/t06_publish/PublishUpdate7Test_gold
cfiehe: Signing is not gone, refer to to `PublishUpdate6Test_gold`.

system/t06_publish/PublishUpdate6Test_gold
cfiehe: This test case was dropped, because `update` now also works for published snapshot repositories. Before the change, it could only be applied to published local repositories.

system/t06_publish/update.py
neolynx: oh, I see, makes sense !

I would not renumber the tests, just leave a gap. some of the tests do not have numbers, but a descriptive name, which i like more somehow...

system/t06_publish/PublishUpdate11Test_gold
```diff
@@ -3,6 +3,5 @@ Generating metadata files and linking package files...
 Finalizing metadata files...
 Signing file 'Release' with gpg, please enter your passphrase when prompted:
 Clearsigning file 'Release' with gpg, please enter your passphrase when prompted:
-Cleaning up prefix "." components main...
```
neolynx, line 6: why is the cleanup gone here ?

system/t06_publish/PublishUpdate7Test_gold
neolynx: ah, the renumbering... I would maybe undo that, makes it hard to review and compare to older runs... what do you think ?

system/t12_api/publish.py
```diff
@@ -1036,3 +1036,207 @@ def check(self):
...
+            json={
+                 "SourceKind": "local",
+                 "Sources": [{"Component": "main", "Name": repo1_name}],
+                 "Signing": DefaultSigningOptions,
+            }
+        )
+
+        # Actual test
+        self.check_equal(
+            self.post(
```
neolynx, line 1228: use post_task here ? then 201 and a task is returned...
otherwise return code is 200 and request is executed synchronously (api returns after completion)

system/t06_publish/AzurePublish2Test_gold
```diff
@@ -3,6 +3,7 @@ Generating metadata files and linking package files...
 Finalizing metadata files...
 Signing file 'Release' with gpg, please enter your passphrase when prompted:
 Clearsigning file 'Release' with gpg, please enter your passphrase when prompted:
-Cleaning up prefix "." components main...
+Cleaning up published repository azure:test1:./maverick...
```
neolynx, line 6: there is a mix of quoted, double quotes, back quoted and not quoted repo/mirror/component references... should we clean that up ?

system/t06_publish/PublishSwitch12Test_gold
```diff
@@ -1,8 +1 @@
-Loading packages...
-Generating metadata files and linking package files...
-Finalizing metadata files...
-Signing file 'Release' with gpg, please enter your passphrase when prompted:
-Clearsigning file 'Release' with gpg, please enter your passphrase when prompted:
-Cleaning up prefix "." components a, c...
-
-Publish for snapshot ./maverick [i386] publishes {a: [snap2]: Created as empty}, {b: [snap2]: Created as empty}, {c: [snap3]: Created as empty} has been successfully switched to new snapshot.
+ERROR: unable to switch: component c does not exist in published repository
```
neolynx, line 1: this is on purpose failing now ?

cfiehe: Yes, I have reverted `switch`, so that the command has the original semantics again. That means you can only change snapshots of existing components and cannot add new components. The test must fail.

system/t06_publish/AzurePublish2Test_gold
cfiehe: I will correct that by using `'` instead of `"`.

cfiehe: It would be nice to quote the `storage:prefix/distribution` as well, but that requires a lot of modifications throughout the whole code. Maybe we can do that when this PR has been merged.

api/publish.go
```diff
@@ -403,3 +519,456 @@ func apiPublishDrop(c *gin.Context) {
...
 	})
 }
+
+// @Summary Add source to staged source list
+// @Description **Add a source to the staged source list**
+// @Description
+// @Description The staged source list exists independently of the current source list of the published repository. It can be modified in multiple steps by adding, removing and updating sources. A source is a tuple of two elements comprising the name of the component and the name of the local repository or snapshot. The staged source list exists as long as it gets discarded via `drop` or applied to the published repository via `update`.
+// @Tags Publish
+// @Param prefix path string true "publishing prefix"
+// @Param distribution path string true "distribution name"
```
neolynx, line 529: Consume json + params

neolynx: somthing like: 
// @Consume json
// @Param request body publishedRepoUpdateSwitchParams true "Parameters"

api/publish.go
```diff
@@ -403,3 +519,456 @@ func apiPublishDrop(c *gin.Context) {
...
+
+// @Summary Set staged source list
+// @Description **Set the staged source list**
+// @Description
+// @Description If the staged source list is known in advance, it can set via this method in a single call. All modifications done before are lost and the staged source list get replaced by the one given in the request body.
+// @Tags Publish
+// @Param prefix path string true "publishing prefix"
+// @Param distribution path string true "distribution name"
+// @Consume json
+// @Param request body publishedRepoUpdateParams true "Parameters"
```
neolynx, line 636: wrong params ?

api/publish.go
```diff
@@ -585,18 +587,22 @@ func apiPublishSourcesCreate(c *gin.Context) {
...
+// @Description
+// @Description Returns added, removed or changed components of snapshots or local repository to be published.
+// @Description
+// @Description The changes will be applied by a subsequent publish update call (See `PUT /api/publish/{prefix}/{distribution}`).
 // @Tags Publish
 // @Param prefix path string true "publishing prefix"
 // @Param distribution path string true "distribution name"
 // @Produce json
-// @Success 200 {array} sourceParams
+// @Success 200 {array} []deb.PublishedRepoRevision
```
cfiehe, line 600: This not correct. `GET` just returns a list of sources. The revision is only used internally.

api/publish.go
```diff
@@ -347,12 +431,28 @@ func apiPublishUpdateSwitch(c *gin.Context) {
...
-	resources = append(resources, string(published.Key()))
-	taskName := fmt.Sprintf("Update published %s (%s): %s", published.SourceKind, strings.Join(updatedComponents, " "), strings.Join(updatedSnapshots, ", "))
+	revision := published.ObtainRevision()
+	sources := revision.Sources
+
+	if published.SourceKind == deb.SourceSnapshot {
+		for _, snapshotInfo := range b.Snapshots {
+			component := snapshotInfo.Component
+			name := snapshotInfo.Name
+			sources[component] = name
```
neolynx, line 441: is `sources` used somewhere ?

deb/publish.go
```diff
@@ -444,20 +654,25 @@ func (p *PublishedRepo) SourceNames() []string {
...
-func (p *PublishedRepo) UpdateLocalRepo(component string) {
+// UpdateLocalRepo inserts/updates local repository source for component
+func (p *PublishedRepo) UpdateLocalRepo(component string, localRepo *LocalRepo) {
 	if p.SourceKind != SourceLocalRepo {
 		panic("not local repo publish")
 	}
 
-	item := p.sourceItems[component]
-	item.packageRefs = item.localRepo.RefList()
+	item, exists := p.sourceItems[component]
```
neolynx, line 663: what is the intention here ? why is the localRepo provided ? can there be multiple ?

api/publish.go
cfiehe: Yes in line 441 :-).

deb/publish.go
cfiehe: Because with this change, it is also possible to change the sources of published local repositories. That was not possible before the change.

api/publish.go
```diff
@@ -347,12 +431,28 @@ func apiPublishUpdateSwitch(c *gin.Context) {
...
+		}
+	}
+
+	resources := []string{string(published.Key())}
+	taskName := fmt.Sprintf("Update published %s repository %s/%s", published.SourceKind, published.StoragePrefix(), published.Distribution)
 	maybeRunTaskInBackground(c, taskName, resources, func(out aptly.Progress, _ *task.Detail) (*task.ProcessReturnValue, error) {
-		err := published.Publish(context.PackagePool(), context, collectionFactory, signer, out, b.ForceOverwrite)
+		result, err := published.Update(collectionFactory, out)
```
neolynx, line 448: maybe run outside of maybeRunTaskInBackground ? this just updates the data structuresm right ?

cfiehe: We can put it outside. I was not sure, what should/must be done outside or inside of a task.

api/publish.go
neolynx: sorry, I don't get it :-)

the sources variable (in 435)  is only assigned (in 441), and not used below, or do I miss something ?

api/publish.go
neolynx: was this part of the logic inside or outside of the background task ?

context: the database will be unlocked, until the background tasks runs, and other tasks usind the same prefix will lock the resource. I think everything modifying should be inside the block. maybe that was not the case before, which could lead to race conditions with parallel operations...

cmd/publish_source_add.go
```diff
@@ -0,0 +1,92 @@
...
+
+The flag -component is mandatory. Use a comma-separated list of components, if
+multiple components should be modified. The number of given components must be
+equal to the number of given sources, e.g.:
+
+	aptly publish add -component=main,contrib wheezy wheezy-main wheezy-contrib
+
+Example:
+
+	$ aptly publish add -component=contrib wheezy ppa wheezy-contrib
```
neolynx, line 81: aptly publish source add?

cmd/publish_source_remove.go
```diff
@@ -0,0 +1,84 @@
...
+		Short:     "remove source from staged source list of published repository",
+		Long: `
+The command removes sources from the staged source list of the published repository.
+
+The flag -component is mandatory. Use a comma-separated list of components, if
+multiple components should be removed, e.g.:
+
+Example:
+
+	$ aptly publish remove -component=contrib,non-free wheezy filesystem:symlink:debian
```
neolynx, line 76: aptly publish source remove ?

cmd/publish_source_update.go
```diff
@@ -0,0 +1,89 @@
...
+		UsageLine: "update <distribution> <source>",
+		Short:     "update source in staged source list of published repository",
+		Long: `
+The command updates sources in the staged source list of the published repository.
+
+The flag -component is mandatory. Use a comma-separated list of components, if
+multiple components should be modified. The number of given components must be
+equal to the number of given sources, e.g.:
+
+	aptly publish update -component=main,contrib wheezy wheezy-main wheezy-contrib
```
neolynx, line 77: aptly publish source update ?

cmd/publish_source_add.go
cfiehe: Fixed: Copy-paste issue 🙈.

cmd/publish_source_remove.go
cfiehe: Fixed: Copy-paste issue 🙈.

cmd/publish_source_update.go
cfiehe: Fixed: Copy-paste issue 🙈.

api/publish.go
cfiehe: `sources` gets defined and set in line 435. The variable is within the loop in line 441.

api/publish.go
cfiehe: The data structure modification was done outside of the task before the change.

api/publish.go
neolynx: ah, it is setting the revision.Sources, adding the components from the Snapshots argument. maybe this deserves a comment ;-)

api/publish.go
neolynx: I think doing this inside is actually fixing a race condition, since go routines run in parallel and multiple requests may overwrite the data structures before the background tasks are started (and the resources are locked).

neolynx:
codecov/patch Successful in 1s — 75.40% of diff hit (target 74.89%)

api/publish.go
```diff
@@ -54,84 +66,163 @@ func slashEscape(path string) string {
...
-	c.JSON(200, result)
+	c.JSON(http.StatusOK, repos)
 }
 
-// POST /publish/:prefix
-func apiPublishRepoOrSnapshot(c *gin.Context) {
+// @Summary Show Published Repository
+// @Description **Get published repository information**
+// @Description
+// @Description Show detailed information of a published repository.
```
neolynx, line 108: reword

api/publish.go
```diff
@@ -384,14 +499,30 @@ func apiPublishUpdateSwitch(c *gin.Context) {
 	})
 }
 
-// DELETE /publish/:prefix/:distribution
+// @Summary Delete Published Repository
+// @Description **Delete a published repository**
+// @Description
+// @Description Delete published repository and clean up files in published directory. Aptly tries to remove as many files belonging to this repository as possible. For example, if no other published repositories share the same prefix, all files inside the prefix will be removed.
```
neolynx, line 505: reword

api/publish.go
```diff
@@ -415,3 +545,486 @@ func apiPublishDrop(c *gin.Context) {
...
+		}
+
+		return &task.ProcessReturnValue{Code: http.StatusOK, Value: revision.SourceList()}, nil
+	})
+}
+
+// @Summary Drop pending changes
+// @Description **Drop pending source component changes of a published repository**
+// @Description
+// @Description Removes all pending changes what would be applied with a subsequent publish update call (See `PUT /api/publish/{prefix}/{distribution}`).
```
neolynx, line 736: add 2nd update call

api/publish.go
```diff
@@ -415,3 +545,486 @@ func apiPublishDrop(c *gin.Context) {
...
+// @Param prefix path string true "publishing prefix"
+// @Param distribution path string true "distribution name"
+// @Consume json
+// @Param request body publishedRepoUpdateParams true "Parameters"
+// @Produce json
+// @Success 200 {object} deb.PublishedRepo
+// @Failure 400 {object} Error "Bad Request"
+// @Failure 404 {object} Error "Published repository/component not found"
+// @Failure 500 {object} Error "Internal Error"
+// @Router /api/publish/{prefix}/{distribution}/update [post]
```
neolynx, line 952: should we use PUT here ? as the other 'update' apis are using PUT... POST would be only for creating the publish point...

cfiehe: This is a special case here, because `/update` has more the semantics of an action call: https://github.com/NationalBankBelgium/REST-API-Design-Guide/wiki/REST-Resources-Actions

I normally use POST in those cases, because it is not a single resource that gets created and the action call may be triggered concurrently.

api/publish.go
```diff
@@ -415,3 +553,488 @@ func apiPublishDrop(c *gin.Context) {
...
+	// Provide index files by hash
+	AcquireByHash *bool `                         json:"AcquireByHash"   example:"false"`
+	// Enable multiple packages with the same filename in different distributions
+	MultiDist *bool `                             json:"MultiDist"       example:"false"`
+}
+
+// @Summary Update Published Repository
+// @Description **Update a published repository**
+// @Description
+// @Description Apply pending changes and republish
```
neolynx, line 949: reword

api/publish.go
```diff
@@ -415,3 +553,488 @@ func apiPublishDrop(c *gin.Context) {
...
+		}
+
+		return &task.ProcessReturnValue{Code: http.StatusOK, Value: gin.H{}}, nil
+	})
+}
+
+// @Summary Update Source Component
+// @Description **Update the source component of a published repository**
+// @Description
+// @Description Publish pending source component changes which were added with `Add/Remove/Replace Source Components`
```
neolynx, line 793: I think that's wrong. we just schedule a source to be published right ?

cmd/publish_source_update.go
```diff
@@ -0,0 +1,89 @@
...
+	return err
+}
+
+func makeCmdPublishSourceUpdate() *commander.Command {
+	cmd := &commander.Command{
+		Run:       aptlyPublishSourceUpdate,
+		UsageLine: "update <distribution> <source>",
+		Short:     "update source in staged source list of published repository",
+		Long: `
+The command updates sources in the staged source list of the published repository.
```
neolynx, line 71: reword

cmd/publish_switch.go
```diff
@@ -111,30 +115,29 @@ func aptlyPublishSwitch(cmd *commander.Command, args []string) error {
...
 func makeCmdPublishSwitch() *commander.Command {
 	cmd := &commander.Command{
 		Run:       aptlyPublishSwitch,
-		UsageLine: "switch <distribution> [[<endpoint>:]<prefix>] <new-snapshot>",
-		Short:     "update published repository by switching to new snapshot",
+		UsageLine: "switch <distribution> [[<endpoint>:]<prefix>] <new-source>",
+		Short:     "update published repository by switching to new source",
 		Long: `
-Command switches in-place published snapshots with new snapshot contents. All
+Command switches in-place published snapshots with new source contents. All
```
neolynx, line 135: reword, this will also apply pending changes now, right ?

cmd/publish_update.go
```diff
@@ -80,14 +76,15 @@ func aptlyPublishUpdate(cmd *commander.Command, args []string) error {
...
+		cleanComponents := make([]string, 0, len(result.UpdatedSources)+len(result.RemovedSources))
+		cleanComponents = append(append(cleanComponents, result.UpdatedComponents()...), result.RemovedComponents()...)
+		err = collectionFactory.PublishedRepoCollection().CleanupPrefixComponentFiles(context, published, cleanComponents, collectionFactory, context.Progress())
 		if err != nil {
 			return fmt.Errorf("unable to update: %s", err)
 		}
 	}
 
-	context.Progress().Printf("\nPublish for local repo %s has been successfully updated.\n", published.String())
+	context.Progress().Printf("\nPublished %s repository %s has been updated successfully.\n", published.SourceKind, published.String())
```
neolynx, line 87: for the documentation below, this will also apply pending changes now, right ?

cmd/publish_source_replace.go
```diff
@@ -0,0 +1,88 @@
...
+	return err
+}
+
+func makeCmdPublishSourceReplace() *commander.Command {
+	cmd := &commander.Command{
+		Run:       aptlyPublishSourceReplace,
+		UsageLine: "replace <distribution> <source>",
+		Short:     "replace staged source list of published repository",
+		Long: `
+The command replaces the staged source list of the published repository.
```
neolynx, line 69: reword

cmd/publish_source_remove.go
```diff
@@ -0,0 +1,84 @@
...
+	return err
+}
+
+func makeCmdPublishSourceRemove() *commander.Command {
+	cmd := &commander.Command{
+		Run:       aptlyPublishSourceRemove,
+		UsageLine: "remove <distribution> [[<endpoint>:]<prefix>] <source>",
+		Short:     "remove source from staged source list of published repository",
+		Long: `
+The command removes sources from the staged source list of the published repository.
```
neolynx, line 69: reword

cmd/publish_source_list.go
```diff
@@ -0,0 +1,89 @@
...
+	return nil
+}
+
+func makeCmdPublishSourceList() *commander.Command {
+	cmd := &commander.Command{
+		Run:       aptlyPublishSourceList,
+		UsageLine: "list <distribution>",
+		Short:     "lists revision of published repository",
+		Long: `
+Command lists sources of a published repository.
```
neolynx, line 76: reword

cmd/publish_source_drop.go
```diff
@@ -0,0 +1,62 @@
...
+	return err
+}
+
+func makeCmdPublishSourceDrop() *commander.Command {
+	cmd := &commander.Command{
+		Run:       aptlyPublishSourceDrop,
+		UsageLine: "drop <distribution>",
+		Short:     "drops staged source changes of published repository",
+		Long: `
+Command drops the staged source changes of the published repository.
```
neolynx, line 50: reword

cmd/publish_source_add.go
```diff
@@ -0,0 +1,92 @@
...
+	return err
+}
+
+func makeCmdPublishSourceAdd() *commander.Command {
+	cmd := &commander.Command{
+		Run:       aptlyPublishSourceAdd,
+		UsageLine: "add <distribution> <source>",
+		Short:     "add source to staged source list of published repository",
+		Long: `
+The command adds sources to the staged source list of the published repository.
```
neolynx, line 71: reword

cmd/publish_switch.go
cfiehe: Hi @neolynx,
no, `publish switch` does not touch the pending changes. This method only exists for reasons of backward compatibility and keeps the former way of changing sources of existing components of published snapshot repositories.

neolynx: will remove publish switch from the 'see also'

cmd/publish_source_replace.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	err = collectionFactory.PublishedRepoCollection().LoadComplete(published, collectionFactory)
+	if err != nil {
+		return fmt.Errorf("unable to add: %s", err)
+	}
+
+	revision := published.ObtainRevision()
+	sources := revision.Sources
+	context.Progress().Printf("Clearing staged source list...\n")
```
neolynx, line 42: do we really need that log ? more of an internal thing, no ?

neolynx: changed to "Replacing source list...\n"

neolynx (approved):

PR #1368, opened 2024-10-09
repo add: improve error message

neolynx:
## Description of the Change

the error message "conflict in package" is due to the package already existing but being different (hash, ..). this change makes the error message more specific.

## Checklist

### DIFF
--- a/deb/list.go
+++ b/deb/list.go
@@ -138,7 +138,7 @@ func (l *PackageList) Add(p *Package) error {
 	existing, ok := l.packages[key]
 	if ok {
 		if !existing.Equals(p) {
-			return &PackageConflictError{fmt.Errorf("conflict in package %s", p)}
+			return &PackageConflictError{fmt.Errorf("package already exists and is different: %s", p)}
 		}
 		return nil
 	}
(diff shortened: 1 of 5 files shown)


deb/list.go
```diff
@@ -138,7 +138,7 @@ func (l *PackageList) Add(p *Package) error {
 	existing, ok := l.packages[key]
 	if ok {
 		if !existing.Equals(p) {
-			return &PackageConflictError{fmt.Errorf("conflict in package %s", p)}
+			return &PackageConflictError{fmt.Errorf("package already exists and is different: %s", p)}
```
an0rak-dev, line 141: Should we say `is different` or `hash mismatch` ? 🤔 

Also, maybe we can add a info or debug log line to display the expected hash (and file size) vs the given one ?

neolynx: unfortunately there are many fields responsible for "different", logging each one seems overkill. 

see deb/package.go:
```
// Equals compares two packages to be identical
func (p *Package) Equals(p2 *Package) bool {
        return p.Name == p2.Name && p.Version == p2.Version && p.SourceArchitecture == p2.SourceArchitecture &&
                p.Architecture == p2.Architecture && p.Source == p2.Source && p.IsSource == p2.IsSource &&
                p.FilesHash == p2.FilesHash
}
```

the term `file already exists and is different` is already used in aptly, so at least it is more consistent...

would you extend this to log all the differing fields ?

szakalboss (approved):

PR #1381, opened 2024-10-22
Cleanup files.PublishedStorage after dropping a publish

aol-nnov:
Fixes: https://github.com/aptly-dev/aptly/issues/198

After dropping files.PublishedStorage there were some leftovers - empty directory tree. If that directory tree is empty, it is now removed.

## Checklist


aol-nnov:
@neolynx here is another annoying thing I'd like to fix in aptly. My publish prefix is bloated with empty directories as I have to create and remove publishes intensively.

neolynx:
oh, yes please ! I found this quite annoying as well...

aol-nnov:
yay, there is an issue for it! https://github.com/aptly-dev/aptly/issues/198

files/public.go
```diff
@@ -119,6 +119,28 @@ func (storage *PublishedStorage) RemoveDirs(path string, progress aptly.Progress
...
+		 ... and so on, up to first non-empty directory
+	*/
+	segments := strings.Split(path, "/")
+
+	for segmentsCount := len(segments); segmentsCount > 0; segmentsCount-- {
+		err := os.Remove(
+			filepath.Join(storage.rootPath,
+				strings.Join(segments[:segmentsCount], "/")))
+
+		if err != nil && strings.Contains(err.Error(), "directory not empty") {
```
neolynx, line 138: could we rather check if it is empty ?

aol-nnov: We could, in theory, but it is a lot more code, than checking an error...

deb/publish.go
```diff
@@ -1323,14 +1324,21 @@ func (collection *PublishedRepoCollection) Remove(publishedStorageProvider aptly
...
-	if !skipCleanup && len(cleanComponents) > 0 {
-		err = collection.CleanupPrefixComponentFiles(repo.Prefix, cleanComponents,
-			publishedStorageProvider.GetPublishedStorage(storage), collectionFactory, progress)
-		if err != nil {
-			if !force {
-				return fmt.Errorf("cleanup failed, use -force-drop to override: %s", err)
+	if !skipCleanup {
+		publishedStorage := publishedStorageProvider.GetPublishedStorage(storage)
+		if len(cleanComponents) > 0 {
+			err = collection.CleanupPrefixComponentFiles(repo.Prefix, cleanComponents,
```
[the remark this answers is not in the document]
aol-nnov: Hmm... Vscode did not change it. Will try manually.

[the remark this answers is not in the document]
aol-nnov: Not much changed, though...

files/public.go
neolynx: i know, but it is also more proper, to not accidentally delete stuff :)

aol-nnov: I do not agree with you. Also, it is algorithmically more complex to perform the check you propose.

In general, the question is: why do you trust one syscall and do not trust another?

neolynx: I think it is a question of taste... ;-) in some industries it is even not allowed to rely on error text, ... not a question of trusting sys calls..

in aptly I would prefer the check before action, we could add a function to utils/utils.go:
```
func isDirEmpty(path string) (bool, error) {
	f, err := os.Open(path)
	if err != nil {
		return false, err
	}
	defer f.Close()

	contents, err := f.Readdirnames(1) // Read only 1 file to check if it's empty
	if err != nil {
		return false, err
	}

	return len(contents) == 0, nil
}
```

and just call this before the os.Remove call. I would not worry about algorithic complexity with a low number of nested directories..

what do you think ?

aol-nnov: Look, even if os.Remove broke and start removing non-empty directories, it won't go beyond publish root. Also, if any of publishes are accidentally lost (which I highly doubt), they could be recreated from corresponding repos.

Anyway, I'm out of office for today and do not have a laptop at hand. Let's get back to it tomorrow.

aol-nnov: Also, if the root of your comment is that it is no good checking error **text**, we could simply bail out on any error. It is perfectly fine also.

neolynx: Look, the file system might also be on a NFS server, and I would like aptly to check before a remove action and have syscalls return errors on real errors. Yes it works both ways, but here in aptly i feel we should do it this way ;-)

Aside from this, there are some tests failing, maybe the tests were wrong, relying on the empty directory ?

aol-nnov: While I'm dealing with failed tests, let's make final round on os.Remove )

I, for one thrive to make code that from one side is bulletproof but from the other is not overloaded with useless actions and expressive enough to read as a poem.

It takes knowing the underlying api well, though.

As for os.Remove, it is wrapper for syscall.Rmdir in case we're removing a directory. Let's dive into various syscall docs now. Follow me.

* [Windows](https://github.com/golang/go/blob/master/src/syscall/syscall_windows.go#L580): it's a [RemoveDirectoryA](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-removedirectorya#return-value) function, which docs say, it "Deletes an existing empty directory."
* [freebsd](https://github.com/golang/go/blob/master/src/syscall/zsyscall_freebsd_amd64.go#L949): it's a rmdir syscall, which [docs say](https://man.freebsd.org/cgi/man.cgi?query=rmdir&apropos=0&sektion=2&manpath=FreeBSD+14.1-RELEASE+and+Ports&arch=default&format=html) "The directory must not have any entries other than `.' and `..'."
* [linux](https://github.com/golang/go/blob/master/src/syscall/syscall_linux.go#L322): https://linux.die.net/man/2/unlinkat -> https://linux.die.net/man/2/rmdir
* [darwin](https://github.com/golang/go/blob/master/src/syscall/zsyscall_darwin_amd64.go#L1340): http://www.polarhome.com/service/man/?qf=RMDIR&af=0&sf=2&of=Darwin&tf=2
* [WASM](https://github.com/golang/go/blob/master/src/syscall/fs_wasip1.go#L697): https://wasix.org/docs/api-reference/wasi/path_remove_directory (read the note at the bottom)

All of them tell the same: it removes an empty directory, returns error otherwise. It's quite normal to try and remove non-empty  directory, handle error, move on. That's what is happening in my code. We're making a syscall and handling the error.

You say: 
>  i feel we should do it this way

Gut feeling is always okay, but you should not follow it blindly!

I strongly believe that review process is a place where we can learn something new and improve our skills every day - that's what is happening here as we type )

If you are still hesitant on the patch set, let's add more reviewers. And be sure, we had internal review, too, and our team came to the conclusion, that checking directory emptiness before os.Remove is excessive, more over, it creates additional points of failure.

aol-nnov:
@neolynx original tests that were failing are now fixed, but my newly added PublishDrop21Test is failing, not sure why...

May be it's time for me to learn something new? This alienish test suite is driving me nuts, frankly...

system/t06_publish/drop.py
```diff
@@ -39,9 +40,33 @@ class PublishDrop2Test(BaseTest):
...
-        self.check_exists('public/ppa/smira/')
+        self.check_not_exists('public/ppa/')
+        self.check_exists('public/')
+
+
+class PublishDrop21Test(BaseTest):
+    """
+    publish drop: under prefix
+    if some publishes have common part of the prefix, only empty parts are cleared
+    (https://github.com/aptly-dev/aptly/pull/1381)
```
neolynx, line 51: no links here, please describe the test behavior like in the other tests

files/public.go
```diff
@@ -119,6 +119,28 @@ func (storage *PublishedStorage) RemoveDirs(path string, progress aptly.Progress
...
+		Say, `path` is "a/b/c/d", let's remove all empty dirs:
+		 storage.rootPath + "a/b/c/d"
+		 storage.rootPath + "a/b/c"
+		 storage.rootPath + "a/b"
+		 ... and so on, up to first non-empty directory (or any other os.Remove error)
+	*/
+	segments := strings.Split(path, "/")
+
+	for segmentsCount := len(segments); segmentsCount > 0; segmentsCount-- {
+		err := os.Remove(
```
neolynx, line 134: I insist on checking for emptiness before remove here :)

also, the the line breaks are not needed, one line please

aol-nnov: > I insist on checking for emptiness

feel free to add it with your commit then

> breaks are not needed

fixed

system/t06_publish/drop.py
aol-nnov: fixed

files/public.go
neolynx: the expectation is that aptly logs if something unexpected happens. and Remove might fail due to various reasons, and we cannot rey on error messages. the only proper way is to check before, bail out if not empty, and log the proper os.Remove error if any, instead of relying on a error to happen. please respect that ;-)

aol-nnov: With all due respect, your approach is error-prone.

https://stackoverflow.com/a/78907172

As said, in Windows actual file removal may be deferred (as stated [in the docs](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-deletefile)), so your userspace check might falsely fail due to that. On the other hand, operations performed in the OS kernel are well aware about all cached modifications made to fs.

As I already stated, feel free adding your implementation with separate commit then.

aol-nnov:
All tests are passing now

neolynx:
> @neolynx original tests that were failing are now fixed

great ! could you explain why they failed ?

> but my newly added PublishDrop21Test is failing, not sure why...
> 
> May be it's time for me to learn something new, too? 😁 This alienish test suite is driving me nuts, frankly...

please do not judge :) the history of all of this is in git, many people worked on that. I think the thing to learn here is how to deal with legacy code and how to slowly improve it and revive the project. so please be respectful and constructive, we are all aware of the state of things, and it took months to get it to a working and halfway reliable state.

aol-nnov:
> could you explain why they failed ?

-gold file was missing for my PublishDrop21Test - copied it from PublishDrop2Test. Not completely sure what it is.

Whole experience of adding tests is cumbersome and non-obvious due to lack of developers documentation. Even some doc links to the testing frameworks would make the whole process more friendly and smooth...

neolynx:
> > could you explain why they failed ?
> 
> -gold file was missing for my PublishDrop21Test - copied it from PublishDrop2Test. Not completely sure what it is.

ok great. in the Makefile there is the CAPTURE flag, you can uncomment, then the tests will create the gold files.

> Whole experience of adding tests is cumbersome and non-obvious due to lack of developers documentation. Even some doc links to the testing frameworks would make the whole process more friendly and smooth...

as said before, we are well aware of that. feel free to contribute. until then let's keep the discussions here technical.
 
> Oh, and original tests were failing because checks relied on empty directories now removed (as you said, the history of all of this is in git :-) )

thanks, that is excellent :-)

files/public.go
```diff
@@ -119,6 +119,26 @@ func (storage *PublishedStorage) RemoveDirs(path string, progress aptly.Progress
...
+		 storage.rootPath + "a/b"
+		 ... and so on, up to first non-empty directory (or any other os.Remove error)
+	*/
+	segments := strings.Split(path, "/")
+
+	for segmentsCount := len(segments); segmentsCount > 0; segmentsCount-- {
+		err := os.Remove(filepath.Join(storage.rootPath, strings.Join(segments[:segmentsCount], "/")))
+
+		if err != nil {
+			break
```
neolynx, line 137: please log the error. we canna see if we have werid permissions like 000 or other behavior.

system/t06_publish/drop.py
neolynx: I still see the link...

aol-nnov: weird... seems git amend set me )

really fixed now

files/public.go
aol-nnov: Okay, log.Printf will suffice? Or should I bring "github.com/rs/zerolog/log" which is used in other files there?

neolynx: in line 117 progress.Printf is used in RemoveDirs. I think we should follow that pattern...

deb/publish.go
```diff
@@ -1323,14 +1324,20 @@ func (collection *PublishedRepoCollection) Remove(publishedStorageProvider aptly
 	collection.list[len(collection.list)-1], collection.list[repoPosition], collection.list =
 		nil, collection.list[len(collection.list)-1], collection.list[:len(collection.list)-1]
 
-	if !skipCleanup && len(cleanComponents) > 0 {
-		err = collection.CleanupPrefixComponentFiles(repo.Prefix, cleanComponents,
-			publishedStorageProvider.GetPublishedStorage(storage), collectionFactory, progress)
-		if err != nil {
-			if !force {
-				return fmt.Errorf("cleanup failed, use -force-drop to override: %s", err)
+	if !skipCleanup {
```
neolynx, line 1327: is this just reformatting ?

files/public.go
aol-nnov: > or return an fmt.Errorf

and what should be done then? It's not the error, that needs any fruther handling.

Implemented it via progress.Printf as you proposed.

deb/publish.go
aol-nnov: No, `len(cleanComponents) > 0` is moved into separate if/else

neolynx: ah, then it is not functional change. please only change what is needed for the PR. this lowers the coverage otherwise

aol-nnov: Please look closer into the patch set. It is needed for the pr...

My added lines https://github.com/aptly-dev/aptly/pull/1381/files#diff-513f82b0090aefbc689e6e0c921004c390c5e8bb8e4240ee133fedcf40cca83dR1338-R1340 should be executed if `skipCleanup` is false but even if `len(cleanComponents)` == 0.

Thus the change.

neolynx: thanks for explaining, that is why i was asking

aol-nnov: Feel free to mark thread as resolved if you're okay with changes.

aol-nnov:
Seems, ci is stuck again...

neolynx:
> Seems, ci is stuck again...

seems we have a segfault:
https://github.com/aptly-dev/aptly/actions/runs/11476603052/job/31936962068?pr=1381#step:8:487

deb/publish.go
neolynx: if you take the publishedStorage before the if block, the nested ifs are not needed, right ?

deb/publish.go
```diff
@@ -1323,14 +1324,20 @@ func (collection *PublishedRepoCollection) Remove(publishedStorageProvider aptly
...
+		if len(cleanComponents) > 0 {
+			err = collection.CleanupPrefixComponentFiles(repo.Prefix, cleanComponents, publishedStorage, collectionFactory, progress)
+			if err != nil {
+				if !force {
+					return fmt.Errorf("cleanup failed, use -force-drop to override: %s", err)
+				}
 			}
 		}
+
+		if localStorage, ok := publishedStorage.(*files.PublishedStorage); ok {
```
neolynx, line 1338: we need to handle also non local endpoints: s3, azure, swift

deb/publish.go
aol-nnov: nope.

See this block on master:

```cpp
if !skipCleanup && len(cleanComponents) > 0 {
		err = collection.CleanupPrefixComponentFiles(repo.Prefix, cleanComponents,
			publishedStorageProvider.GetPublishedStorage(storage), collectionFactory, progress)
		if err != nil {
			if !force {
				return fmt.Errorf("cleanup failed, use -force-drop to override: %s", err)
			}
		}
	}
```

Now in this PR:

```
	if !skipCleanup {
		publishedStorage := publishedStorageProvider.GetPublishedStorage(storage)
		if len(cleanComponents) > 0 {
			err = collection.CleanupPrefixComponentFiles(repo.Prefix, cleanComponents, publishedStorage, collectionFactory, progress)
			if err != nil {
				if !force {
					return fmt.Errorf("cleanup failed, use -force-drop to override: %s", err)
				}
			}
		}

		if localStorage, ok := publishedStorage.(*files.PublishedStorage); ok {
			localStorage.CleanupPublishTree(repo.Prefix, progress)
		}
	}
```

original lines moved into separate `if len(cleanComponents) > 0` block. My additions are next to it, not inside.

Both parts are enclossed in `if !skipCleanup`

neolynx: got it

deb/publish.go
aol-nnov: s3 does not work as local fs - there are no such things as "directories". That's why I opted to `files.PublishedStorage` method and not `aptly.PublishedStorage` interface.

aol-nnov: Proof link: https://www.reddit.com/r/aws/comments/vizzqr/uploading_empty_folders_as_objects_to_aws_s3/

neolynx: ok, great. what about the others ?

sorry for being picky, I am asking bcs new aptly features should be complete across api/commands/endpoints/databases... just trying to keep things consistent from now on :)

aol-nnov:
> seems we have a segfault

Heh,  that's because nil is passed instead of `aptly.Progress` in tests. Very handy 😒

deb/publish.go
aol-nnov: Having `func (storage *PublishedStorage) MkDir(_ string) error {` implemented only for local file storage leads me to conclusion that s3, azure and swift do not have such thing as "empty directory"

https://github.com/search?q=repo%3Aaptly-dev%2Faptly+*PublishedStorage%29+MkDir&type=code

files/public.go
neolynx: on a (misconfigured) remote file server, there might be permission issues when removing directories. such errors should be logged... this is why I would still like to have a check for emptyness, bcs then we do not get an error. 

do we have a check where a directory is not removed bcs not empty ?

another little thing: progress logs in aptly do not have punctiations, could you make the log similar to the others ?

aol-nnov:
Now, seems `-gold`-en outputs do not match new reality 😂

neolynx:
> Now, seems `-gold`-en outputs do not match new reality 😂

it needs that non empty check before remove... ;-)

files/public.go
aol-nnov: Oh, too much moving parts here..

> should be logged

but strings printed by aptly.Progress are seen only during cli operation

> remote file server

In api mode all logging is handled by gin or zerolog, I do not know for sure, but plain fmt.Println from gin controller in aptly outputs nothing to stdout.

Your intentions regarding localStorage.CleanupPublishTree() "errors" as you call them is not clear.

Imagine, we will return an error object from it, handle it in `func (collection *PublishedRepoCollection) Remove()` like

```go
		if localStorage, ok := publishedStorage.(*files.PublishedStorage); ok {
			if err := localStorage.CleanupPublishTree(repo.Prefix, progress); err != nil {
				return fmt.Errorf("")
			}
		}
	}

	batch := collection.db.CreateBatch()
	batch.Delete(repo.Key())

	for _, component := range repo.Components() {
		batch.Delete(repo.RefKey(component))
	}

	return batch.Write()
}
```

But is that **really** an error of a scale, that we should terminate the whole process and to not update database at this point?!

Keep in mind, that by the time localStorage.CleanupPublishTree() is executed, all files concerned the publish being `Remove()`-ed are long ago gone. So, if we terminate the execution of Remove() at this point, we'll find ourselves in an inconsistent state!

To my understanding, silently leaving directories intact if we can not remove them is the best bet in this situation. 
Okay, we'll log this in cli via `progress`, but how about api mode?

Also, as I mentioned earlier, I was unable log anything to console by plain printf from gin controller.

aol-nnov:
Also, something wrong with docker tests:

```
$ make docker-image
[+] Building 310.3s (21/21) FINISHED
$ make docker-system-test
usermod: UID '0' already exists
make: *** [docker-system-test] Error 4
$ _
```

So, for me there is no simple way to update test now, unfortunately.

aol-nnov:
> it needs that non empty check before remove... ;-)

You know what, my experience contributing in this particular PR is drastically falling down. Second day in a row we're chewing through tons of comments in a changeset that barely changes nothing, docs on testsuite are missing, tooling does not work and you're making jokes on me...

That all really makes me sad.

neolynx:
> Also, something wrong with docker tests:

the docker-wrapper sets the user to the one of the git repo directory, to not write files as root. it seems your git dir is owned by root. changing to user owner would solve this in that case.

aol-nnov:
> it seems your git dir is owned by root

not true

```
$ ls -lad ../aptly/
drwxr-xr-x@ 46 aol  staff  1472 Oct 22 17:22 ../aptly/
```

aol-nnov:
Re: better tooling https://github.com/aptly-dev/aptly/pull/1384

It might be worth merging it first.

files/public.go
aol-nnov: Notice from `localStorage.CleanupPublishTree` is being logged as per your request and is visible in cli and api modes.

aol-nnov:
Any chance there is some substitute for a timestamp in `_gold`-files?

```patch
-{"level":"debug","time":"2024-10-29T12:23:46Z","message":"CleanupPublishTree: remove /home/runner/.aptly/public/ppa: directory not empty. '/home/runner/.aptly/public/ppa' not removed\n"}
+{"level":"debug","time":"2024-10-30T06:44:38Z","message":"CleanupPublishTree: remove /home/runner/.aptly/public/ppa: directory not empty. '/home/runner/.aptly/public/ppa' not removed\n"}
```

Now test is failing due to it

aol-nnov:
Rebased to the tip of `master`, conflicts resolved.

system/t06_publish/PublishDrop21Test_gold
```diff
@@ -0,0 +1,5 @@
+Removing ${HOME}/.aptly/public/ppa/smira/dists...
+Removing ${HOME}/.aptly/public/ppa/smira/pool...
+{"level":"debug","time":"now","message":"CleanupPublishTree: remove ${HOME}/.aptly/public/ppa: directory not empty. '${HOME}/.aptly/public/ppa' not removed\n"}
```
neolynx, line 3: we need to use normal logging here, no json to console...

neolynx:
Dear @aol-nnov,

I was by no means making jokes on you, i was merely a wink to point out what is expected here. Your skills and contributions are very welcome, also for documentation and all the other stuff which is lagging behind !

I talked to some of the other maintainers, here are our thougths:
- Aptly does not have the authority over the prefix path, as it is most likely used in a webserver directory structure, which might be managed by other tools 
- Changing the default behavior to remove the tree could result in 404 instead of a empty directory index on the webserver
- Proper logging of unexpected error is required, as aptly logs are crucial for reproducing issues, also expected errors are prevented / not logged
- there is also `deb/publish.go:func (p *PublishedRepo) RemoveFiles(publishedStorageProvider aptly.PublishedStorageProvider, removePrefix bool,` - the removePrefix might be doing something similar, maybe this can be combined?
- Issue comments are marked as resolved by the maintainers. Even if the concern might feel unreasonable or not to your liking.

To move forward here, we suggest the following changes:
- make a cleanupTree argument to the drop calls (api, cmd), cleanup up the tree only if set to true and skipCleanup is false
- in CleanupPublishTree, check if the directory is empty and return if true, otherwise remove directory. return the error and log in the caller
- original test should prove that prefix is not removed, new tests prove that empty dirs are removed, non-empty ones not (without a error log)
- if we wanna be fancy, a test with a parent directory without write permissions would prove that the error is logged

Please understand that aptly is a legacy project with many maintainers and contributors over the years. It took considerable effort to revive it and bring it to the state it is now. As maintainers, our main concern is to not break any of potentially thousands of installations, and if something happens, we have proper logs. And this has some implications on how we can implement things. 

@cfiehe, @szakalboss, @dario-gallucci please add if I forgot something

aol-nnov:
Dear @neolynx,

First of all, I'd like to point that all stated below is not personal offence, but a feedback for the project maintainer handling incoming changes.

Overall experience during contributing to this miserly PR was cumbersome and disappointing. Starting from constantly changing and somewhat controversial requirements (I'm not even asking now how would checking directory for the emptiness before removal will aid 404s on missing dirs and such and such) up to never-ending discussions which made me feel totally not understood, and to suddenly-disappearing-reviewer right in the middle of threads without any notice on coming-back-or-not and many other obstacles along the way...

As we know, all good things eventually come to their end. So do my resources, that I'm able to invest into this PR.

Feel free to implement this change exactly "the way you feel", as all technical reasons I brought to you were discarded or ignored by you along the way. Also feel free to take the credits for this change - I do not mind.

Our enterprise is heavily using aptly in production for over a decade now (and contributing to it, as well), and we were forced to operate on our fork for a few times already, rather than upstrean version due to abandonedness or the project, beurocracy of accepting patches or mantainer's irresponsiveness.

Sadly, it seems, we're on the fork again. We need to mind our business, not to prove something on the Internet.

Good luck and thanks for all the fish. (Making it clear: I'm no longer able to work on this PR.)

PR #1384, opened 2024-10-24
Improve dev tooling

aol-nnov:
* Add vscode recommended config and plugins
* Update docker machinery
* Simplify Makefile, eliminate the need of editing to enable CAPTURE for new tests. Now simply `make (docker-)system-test CAPTURE=yes`

Previously created docker stuff did not work (at
least in Dokcer Desktop for Mac). Let's try and
create something more generalzied. Yet, changes
were not tested on WSL.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,13 @@
+{
+    "recommendations": [
+        "aaron-bond.better-comments",
+        "jeff-hykin.better-go-syntax",
+        "ms-azuretools.vscode-docker",
+        "golang.go",
+        "stkb.rewrap",
+        "gruntfuggly.todo-tree",
+        "ms-python.vscode-pylance",
+        "ms-python.python",
+        "ms-python.debugpy"
+    ]
+}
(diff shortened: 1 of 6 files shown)


aol-nnov:
`usermod -u `stat -c %u /work/src` aptly` did not work as expected on Docker Desktop for Mac (attached volume mount point is owned by root). Which leads to

```
$ make docker-image
[+] Building 310.3s (21/21) FINISHED
$ make docker-system-test
usermod: UID '0' already exists
make: *** [docker-system-test] Error 4
$ _
```

So, let's incorporate developer's UID and GID into image at build time.

N.B.: image is now "developer machine"-specific. Which is of no harm, I suppose, as it is only used for local development.

neolynx:
> N.B.: image is now "developer machine"-specific. Which is of no harm, I suppose, as it is only used for local development.

the reason for the docker-wrapper is to have the image not user specific. someone might also use multiple users for testing, which would use the same image.

neolynx:
it would be interesting why on MacOS the owner is root, and if all files are...

neolynx:
unfortunately, this also removes the tab completion for the docker make targets and the info from make help :(

neolynx:
please reopen with more info about mac os volume permissions

Issue #1385, opened 2024-10-28
-filter-with-deps flag isn't including dependencies


otherjason:
## Detailed Description

I am trying to create a mirror from a portion of the repository found at https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2204/x86_64/. This repo contains many releases of NVIDIA's CUDA toolkit, which are large in size. I'm trying to create a mirror of just the latest release.

There is a single package, `cuda-12-6`, that can be used as a top-level installation target; it depends on the other packages in the repo that actually contain the functionality. The output of `dpkg -I` for this package is:

```
$ dpkg -I cuda-12-6_12.6.2-1_amd64.deb
 new Debian package, version 2.0.
 size 2522 bytes: control archive=584 bytes.
     492 bytes,    12 lines      control              
      78 bytes,     1 lines      md5sums              
 Package: cuda-12-6
 Version: 12.6.2-1
 Architecture: amd64
 Maintainer: cudatools <cudatools@nvidia.com>
 Installed-Size: 7
 Depends: cuda-runtime-12-6 (>= 12.6.2), cuda-toolkit-12-6 (>= 12.6.2), cuda-demo-suite-12-6 (>= 12.6.77)
 Section: multiverse/devel
 Priority: optional
 Description: CUDA 12.6 meta-package
  Meta-package containing all the available packages required for native CUDA
  development. Contains the toolkit, samples, driver and documentation. Locked
  at CUDA Toolkit version 12.6.
```

I tried to use this meta-package to create a mirror containing this package only and all of its dependencies. However, when I do so, `aptly` only downloads the meta-package and does not download the dependencies. Here's what I got:

```
$ aptly mirror create -filter "cuda-12-6 (= 12.6.2-1)" -filter-with-deps cuda-12.6.2-ubuntu-22.04 https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2204/x86_64/ /
Downloading: https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2204/x86_64//InRelease
gpgv: Signature made Fri 25 Oct 2024 03:39:17 PM EDT
gpgv:                using RSA key A4B469963BF863CC
gpgv: Good signature from "cudatools <cudatools@nvidia.com>"

Mirror [cuda-12.6.2-ubuntu-22.04]: https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2204/x86_64/ .// successfully added.
[...]

Issue #1378, opened 2024-10-21
False positive in ImportPackageFiles when adding .deb files that have different metadata and name but same contents


ferrreo:
## Detailed Description
If you have a bunch of packages that differ only in package name and depends then when you try add them to your repo via repo add and you have forceReplace set then one of them gets booted out even though they do not actually match or conflict.

This is due to ImportPackageFiles inside the force replace block it is doing a list.search but this is not a thorough as package.Equals, if you add check that matches package.Equals minus the filehash in the loop ranging over "conflictingFiles" then the problem goes away.

Ideally the fix would be in list.search or it's params.

## Context
This caused some serious confusion when packages went "missing" from our Distro repo. 
This is a pretty common usecase when it comes to metapackages, as generally they have no content and just differ by their depends.

## Possible Implementation
I fixed it locally with adding a check to package.SoftEquals in the looping over conflictingFiles and only marking as conflicted if the check passes. SoftEquals is just package.Equals minus the fileshash check.

## Your Environment
Not really environment specific but I can provided debs or an example control file if needed.

neolynx:
Hi !

thanks for the analysis :-)

would you mind opening a PR, so we can have a look ?

neolynx:
is there an easy way to reproduce this ?

ferrreo:
Apologies, been super busy since I raised this. To replicate try and add two debs that have different names and different deps in the control (common for virtual packages) at the same time (e.g repo add a dir with them both in).

Here are two of our debs we had issues with:

https://ppa.pika-os.com/pool/cockatiel/m/mesa/mesa-stable_24.2.6-101pika1_amd64.deb
https://ppa.pika-os.com/pool/cockatiel/m/mesa/mesa-stable-no-march_24.2.6-101pika1_amd64.deb

I can provide the control file too if needed.

The code I used to "fix" it:

In import.go:

`		if forceReplace {
			conflictingPackages := list.Search(Dependency{Pkg: p.Name, Version: p.Version, Relation: VersionEqual, Architecture: p.Architecture}, true)
			for _, cp := range conflictingPackages {
				if !cp.SoftEquals(p) {
					reporter.Warning("%s marked as conflicted but didn't match", p)
					continue
				}
				reporter.Removed("%s removed due to conflict with package being added", cp)
				list.Remove(cp)
			}
		}
`

package.go

`func (p *Package) SoftEquals(p2 *Package) bool {
[...]

PR #1386, opened 2024-11-04
Fix/add provided package

neolynx:
Fixes https://github.com/aptly-dev/aptly/issues/1378

## Description of the Change

Since https://github.com/aptly-dev/aptly/commit/ab18d4835bc726b4f65db69b87c03aea72d63e07 list.Search() is also returning provided packages. This is not wanted in all scenarios, for example then adding packages to a repository or in a snapshot pull. 

## Checklist

- [ ] unit-test added (if change is algorithm)
- [ ] functional test added/updated (if change is functional)

PR #1387, opened 2024-11-05
package.go: Fix bug in providesDependency

5hir0kur0:
While trying to debug #1378, I noticed that my new version of `providesDependency` is probably wrong.
In the case when the `Provides` entry does not specify a version, calling `providesDependency` with `VersionEqual` as the relation would return `false`, IDK if this is the correct behavior. I think the correct behavior would be to use the version specified in the `Version` field in this case.
Also, it was only considering the `Relation` of the `Provides:` entry instead of the relation of the actual dependency it seems.

Could you please check this?

## Requirements

## Checklist

- [ ] documentation updated

### DIFF
--- a/deb/package.go
+++ b/deb/package.go
@@ -362,24 +362,22 @@ func (p *Package) providesDependency(dep Dependency) (bool, error) {
 		if err != nil {
 			errs = append(errs, err)
 		}
-		// The only relation allowed here is `=`.
-		// > The relations allowed are [...]. The exception is the Provides field, for which only = is allowed.
-		// > [...]
-		// > A Provides field may contain version numbers, and such a version number will be considered when
-		// > considering a dependency on or conflict with the virtual package name.
-		// -- https://www.debian.org/doc/debian-policy/ch-relationships.html
-		switch providedDep.Relation {
-		case VersionDontCare:
-			if providedDep.Pkg == dep.Pkg {
-				return true, nil
-			}
-		case VersionEqual:
-			providedVersion := providedDep.Version
-			if providedDep.Pkg == dep.Pkg && versionSatisfiesDependency(providedVersion, dep) {
-				return true, nil
-			}
-		default:
+		if providedDep.Relation != VersionEqual && providedDep.Relation != VersionDontCare {
+			// The only relation allowed here is `=`.
+			// > The relations allowed are [...]. The exception is the Provides field, for which only = is allowed.
+			// > [...]
+			// > A Provides field may contain version numbers, and such a version number will be considered when
+			// > considering a dependency on or conflict with the virtual package name.
+			// -- https://www.debian.org/doc/debian-policy/ch-relationships.html
 			errs = append(errs, fmt.Errorf("unsupported relation in Provides: %s", providedDep.String()))
+			continue
+		}
+		providedVersion := providedDep.Version
+		if providedVersion == "" {
+			providedVersion = p.Version
+		}
+		if providedDep.Pkg == dep.Pkg && versionSatisfiesDependency(providedVersion, dep) {
+			return true, nil
 		}
 	}
 	return false, JoinErrors(errs...)
--- a/system/t05_snapshot/VerifySnapshot7Test_gold
+++ b/system/t05_snapshot/VerifySnapshot7Test_gold
@@ -1,6 +1,6 @@
 Loading packages...
 Verifying...
-Missing dependencies (625):
+Missing dependencies (635):
   abrowser [amd64]
   abrowser [i386]
   apache [amd64]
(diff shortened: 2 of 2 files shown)


neolynx:
hi @5hir0kur0,

thanks for the analysis !

do you think this is still needed if we merge https://github.com/aptly-dev/aptly/pull/1386 ?

5hir0kur0:
> hi @5hir0kur0,
> 
> thanks for the analysis !
> 
> do you think this is still needed if we merge #1386 ?

Yes, it's an unrelated bug. As you can see it also affects the output of some system tests, though I think/hope that the new output is more correct than before.

neolynx:
oh ! I did not fully understand that, thanks :)

looking at the errors, i thought they look somewhat familiar: https://github.com/aptly-dev/aptly/commit/95915480a0e24a7c79f61ed57a745fc46202fef2#diff-d92c1b645b188fef92584e65238270ef2e01e3690ce9d7614d6f42abe6592d73R3

The VerifySnapshot7Test originally had 677, then this was reduced to 625. this change here brings some dependencies back, for example  git-core, libtest-harness-perl, python-codespeak-lib, xpdf-utils, ...

could you verify that those packages are indeed missing dependencies, while the other ones not (for example fvwm-gnome, grep-dctrl)?

neolynx:
I don't seem to have write access, could you provide it please ?

I tried to rebase on master :)

Also, I hoped that this would maybe fix https://github.com/aptly-dev/aptly/issues/1385, but unfortunately not...

5hir0kur0:
> could you verify that those packages are indeed missing dependencies, while the other ones not (for example fvwm-gnome, grep-dctrl)?

The re-added packages do appear to be actually missing dependencies. 
I added some debug output to the VerifyDependencies function and ran the command from the test.
```
'git-core' is needed by 'guilt'
  -> dependencies of guilt: [git-core (>= 1.4) | git (>= 1:1.7) git-core (<= 1:1.7.0.4-1) | git (<< 1:1.8)]
'xpdf-utils' is needed by 'impressive'
  -> dependencies of impressive: [python python-support (>= 0.90.0) python-opengl python-pygame python-imaging poppler-utils | xpdf-reader (<< 3.02-2) | xpdf-utils (>= 3.02-2) perl]
'libhaml-ruby' is needed by 'libplack-middleware-file-sass-perl'
  -> dependencies of libplack-middleware-file-sass-perl: [perl libplack-perl ruby-sass | libhaml-ruby (<< 3.1) | libtext-sass-perl]
'libtest-harness-perl' is needed by 'libtap-parser-sourcehandler-pgtap-perl'
  -> dependencies of libtap-parser-sourcehandler-pgtap-perl: [perl perl-modules (>= 5.14) | libtest-harness-perl (= 3.23-1)]
'python-codespeak-lib' is needed by 'python-gevent'
  -> dependencies of python-gevent: [libc6 (>= 2.3.6-6~) libevent-2.0-5 (>= 2.0.10-stable) python (<< 2.8) python (>= 2.6) python-support (>= 0.90.0) python-greenlet | python-codespeak-lib (<< 1.0)]
``` 

E.g., `xpdf-utils` is needed in `>= 3.02-2` but is provided in version `0.18.4-6`.

```
provides of 'poppler-utils' (version 0.18.4-6): [pdftohtml xpdf-utils]
```

The previous code before this fix just ignored the version of the provided packages unless the `Provides:` entry explicitly had a version specified with the `(= ...)` syntax.

I'm not 100% sure with `git-core` since it is provided by the `git` package (`1:1.7.10.4-1+wheezy1`) which I guess is then counted as missing because it's not `<= 1:1.7.0.4-1`

```
provides of 'git' (version 1:1.7.10.4-1+wheezy1): [git-completion git-core]
```

```shell
$ dpkg --compare-versions 1:1.7.10.4-1+wheezy1 le 1:1.7.0.4-1; echo $?
1
```

> could you verify that those packages are indeed missing dependencies, while the other ones not (for example fvwm-gnome, grep-dctrl)?

`fvwm-gnome` is not missing, because it's provided by `fvwm` 

```
dependencies of 'fvwm-crystal': [python python-support (>= 0.90.0) fvwm (>= 1:2.5.13) | fvwm-gnome (>= 1:2.5.13) rxvt-unicode | x-terminal-emulator imagemagick trayer | stalonetray habak | hsetroot]
provides of 'fvwm' (version 1:2.5.30.ds-1.1): [fvwm-gnome x-window-manager]
```
`grep-dctrl` is also provided:

```
provides of 'dctrl-tools' (version 2.22.2): [grep-dctrl]
```
I also checked a few others, though not all because IDK how to automate it.
Here's the list of all `Provides:` entries and dependencies in `snap1` from the system test for reference ([provided-packages-and-dependencies.txt](https://github.com/user-attachments/files/17679038/provided-packages-and-dependencies.txt)).

neolynx:
wanted to add a fix for the test, but cannot push:

error: Authentication error: Authentication required: You must have push access to verify locks
error: failed to push some refs to 'github.com:5hir0kur0/aptly.git'

neolynx:
Thanks @5hir0kur0 for digging into the depths of this and sounds explanations !

neolynx (approved):

PR #1388, opened 2024-11-08
do not set empty mirror architectures for flat mirrors

neolynx:
Fixes https://github.com/aptly-dev/aptly/issues/1385

## Description of the Change

when creating a mirror of a flat repository, aptly sets the architectures to nil. This prevents filtering with dependencies, because this is done by looping through the architectures.

This change fixes this by forcing all mirrors to have an architecture (and those which don't require the architectures specified when creating the mirror). If available, the architectures will be set from the Release file. 

## Checklist


iofq (approved):
LGTM!

PR #1389, opened 2024-11-08
Remove lfs

neolynx:
Fixes CI

## Description of the Change

Turns out git lfs was a bad idea, due to github bandwith limitations... (we'll have to find a different way to save AWS costs)...


neolynx:
Merging bcs blocks pipeline

PR #1390, opened 2024-11-08
Make HTTP server wait for tasks before shutdown

### LINKED ISSUE #1335: Aptly DB deadlock when running in container as PID 1
Aptly stores and uses a "worker PID" in the DB to check for running background processes and determine whether a repo is still "locked" by a bg process or is safe for access by another process. When running in a container using `api serve` command, this PID is set to 1. If the container restarts, the new Aptly will see PID 1 as the owner, check for a running PID 1 (which exists, both the old and new Aptly are PID 1), and maintain that there's a lock on that repo. However, the thread running the update process is long gone, since the container was restarted. Thus, the lock is held forever. 

## Context
Without this change, Aptly becomes kinda unstable as an API-only tool running in a container

## Possible Implementation
I think aptly should be doing better cleanup of this worker pid on SIGTERM. I debugged this to https://github.com/aptly-dev/aptly/blob/master/api/mirror.go#L396 which looks like it does what we want, it just does not run on context cancel and context cancel does not catch SIGTERM.

## Your Environment
To reproduce: 
* In a container: `aptly api serve`
* curl -X POST -H "Content-Type: application/json" --data '{"Name": "ubu", "Distribution": "jammy", "ArchiveURL": "http://us.archive.ubuntu.com/ubuntu"}' localhost:8080/api/mirrors?_async=1
* curl -X PUT localhost:8080/api/mirrors/ubu2?_async=1
* Restart container while PUT/update is running
* curl -X DELETE localhost:8080/api/mirrors/ubu2?_async=1
* observe deadlock as it looks for the old PID 1 to complete

iofq:
Fixes #1335 

## Description of the Change

This code change prevents SIGINT/SIGTERM signals from stopping the Aptly API server before it has cleaned up any resource locks in the database. Normally for CLI aptly commands, it would clear these locks if it detects the PID that had the previous lock is no longer running. However, if Aptly is running as an API server in a container, it always runs as PID 1, which means the PID no longer works as a unique identifier to detect what is running at any given time, leaving these locks indefinitely.

Even with this MR, this issue can still present itself if an Aptly container is SIGKILLed (docker rm -f), however that is a more difficult problem to solve that would likely take some significant rearchitecting (since the API was tacked on to aptly after the fact). 

## Checklist

### DIFF
--- a/cmd/api_serve.go
+++ b/cmd/api_serve.go
@@ -65,6 +65,7 @@ func aptlyAPIServe(cmd *commander.Command, args []string) error {
 	signal.Notify(sigchan, syscall.SIGINT, syscall.SIGTERM)
 	go (func() {
 		if _, ok := <-sigchan; ok {
+			context.TaskList().Wait()
 			server.Shutdown(stdcontext.Background())
 		}
 	})()
--- a/cmd/api_serve.go
+++ b/cmd/api_serve.go
@@ -65,6 +65,7 @@ func aptlyAPIServe(cmd *commander.Command, args []string) error {
 	signal.Notify(sigchan, syscall.SIGINT, syscall.SIGTERM)
 	go (func() {
 		if _, ok := <-sigchan; ok {
+			fmt.Printf("\nShutdown signal received, waiting for background tasks...\n")
 			context.TaskList().Wait()
 			server.Shutdown(stdcontext.Background())
 		}
--- a/context/context.go
+++ b/context/context.go
@@ -13,6 +13,7 @@ import (
 	"runtime/pprof"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/aptly-dev/aptly/aptly"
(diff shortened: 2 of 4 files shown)


cmd/api_serve.go
```diff
@@ -65,6 +65,7 @@ func aptlyAPIServe(cmd *commander.Command, args []string) error {
 	signal.Notify(sigchan, syscall.SIGINT, syscall.SIGTERM)
 	go (func() {
 		if _, ok := <-sigchan; ok {
+			context.TaskList().Wait()
```
neolynx, line 68: would be great to add some logging, something like "signal received, waiting for tasks to complete". this would help debugging I think...

neolynx:
excellent work ! thanks a lot :)

I ran a coverage build: https://github.com/aptly-dev/aptly/pull/1391 - 60.00% of diff hit (target 74.86%), which is good enough, but I was wondering if we can somehow test the Ctrl-C automatically. There is already a test for parallel operation with threads... do you think it would make sense to try to test this ?

iofq:
> excellent work ! thanks a lot :)
> 
> I ran a coverage build: #1391 - 60.00% of diff hit (target 74.86%), which is good enough, but I was wondering if we can somehow test the Ctrl-C automatically. There is already a test for parallel operation with threads... do you think it would make sense to try to test this ?

I added a unit test to cover the added lines. As far as a system test, I messed a bit but could not figure out how to cleanly shut down and re-init the test harness mid-test easily

neolynx (approved):
thanks !

Issue #1290, opened 2024-05-11
Azure Storage Blob SDK for Go DEPRECATED


scentoni:
The library https://github.com/Azure/azure-storage-blob-go/ that aptly uses to deploy packages to Azure is deprecated and will be retired 2024-09-13. As described there, it should be replaced with https://github.com/Azure/azure-sdk-for-go before that date.

Comparing the old azblob SDK documentation https://pkg.go.dev/github.com/Azure/azure-storage-blob-go/azblob and the new azblob SDK documentation https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/storage/azblob, it does not appear to be a drop-in replacement.

SingulariteeSplunk:
@neolynx , @chuan , @randombenj , @lbolla , @andrewshadura - Apologies for all the tags, but wasn't sure who would be most appropriate. It looks like we have an incoming breaking change soon for the newly added Azure support. Does anyone have cycles to take a look?

neolynx:
Thanks for reporting !!

as I don't really have azure storage access, I would need some help with that.. 

would you be able to prepare a pull request ?

plane-wave:
@neolynx Hi there! Do you still need access to an Azure storage instance to test this? Or I read elsewhere that you would like to stay on a particular Go version.

neolynx:
the go version has been updated to 1.22.7 ! would that be suitable for the new Azure SDK ?

yes, if you could help with testing Azure (i.e. have a test account) that would be great. I had a free tier, but it expired..

neolynx:
also, aptly uses https://github.com/potatoqualitee/azuright 1.1 for testing azure. but there seems to be no newer version. it would be greate somehow if the new sdk could be emulated for testing. 

do you think this is already possible with 1.1 ?

neolynx:
I migrated to the new azure sdk, kinda, it compiles but unit tests now fail (on azurite)...

neolynx:
test are passing now ! azurite was used for testing

@plane-wave would you be able to compile and test the branch with a real Azure ? especially with using a publish prefix, not sure it is handled correctly everywhere...

@scentoni, @SingulariteeSplunk would you be able to test as well ?

andrewshadura:
@em-, maybe you can test this?

neolynx:
branch is: fix/azure-sdk-migration

It would be great to test with big files, symlinks (link from pool, acquireByHash), removing files...
[...]

PR #1392, opened 2024-11-11
Migrate to  azure-sdk-for-go 

neolynx:
Fixes https://github.com/aptly-dev/aptly/issues/1290

## Description of the Change

- migrate to new azure sdk

## Checklist

- [x] FIXMEs, MD5 upload, ...


azure/public.go
```diff
@@ -114,9 +115,9 @@ func (storage *PublishedStorage) LinkFromPool(publishedPrefix, publishedRelPath,
 	sourcePath string, sourceChecksums utils.ChecksumInfo, force bool) error {
 
 	relFilePath := filepath.Join(publishedRelPath, fileName)
-	// prefixRelFilePath := filepath.Join(publishedPrefix, relFilePath)
-	// FIXME: check how to integrate publishedPrefix:
-	poolPath := storage.az.blobPath(fileName)
+	prefixRelFilePath := filepath.Join(publishedPrefix, relFilePath)
+	// FIXME: check how to integrate publishedPrefix: done?
```
neolynx, line 119: use storage.az.blobPath

azure/public.go
```diff
@@ -236,12 +246,12 @@ func (storage *PublishedStorage) internalCopyOrMoveBlob(src, dst string, metadat
...
 // RenameFile renames (moves) file
 func (storage *PublishedStorage) RenameFile(oldName, newName string) error {
-	return storage.internalCopyOrMoveBlob(oldName, newName, nil, true /* move */)
+	return storage.internalCopyOrMoveBlob(oldName, newName, "", true /* move */)
 }
 
 // SymLink creates a copy of src file and adds link information as meta data
 func (storage *PublishedStorage) SymLink(src string, dst string) error {
-	return storage.internalCopyOrMoveBlob(src, dst, azblob.Metadata{"SymLink": src}, false /* move */)
+	return storage.internalCopyOrMoveBlob(src, dst, "SymLink", false /* do not remove src */)
```
neolynx, line 254: fixme

iofq (changes requested):
Need to update `debian/control` with the new azure go package

neolynx:
> Need to update `debian/control` with the new azure go package

I tried that, unfortunately the azure sdk package is lacking some modules in the latest version on debian, so this has to wait.
I am also synchronizing with the official debian packaging, and will ask if the missing go modules (mainly azure, swaggo, ..) could be provided for trixie.
The github pipeline is not building with debian dependencies, so the debian/control does not matter there...

iofq:
> > Need to update `debian/control` with the new azure go package
> 
> I tried that, unfortunately the azure sdk package is lacking some modules in the latest version on debian, so this has to wait. I am also synchronizing with the official debian packaging, and will ask if the missing go modules (mainly azure, swaggo, ..) could be provided for trixie. The github pipeline is not building with debian dependencies, so the debian/control does not matter there...

gotcha so this must just be for building from source (deb-src)?

iofq (approved):

neolynx:
> > > Need to update `debian/control` with the new azure go package
> > 
> > 
> > I tried that, unfortunately the azure sdk package is lacking some modules in the latest version on debian, so this has to wait. I am also synchronizing with the official debian packaging, and will ask if the missing go modules (mainly azure, swaggo, ..) could be provided for trixie. The github pipeline is not building with debian dependencies, so the debian/control does not matter there...
> 
> gotcha so this must just be for building from source (deb-src)?

yes, this is just for debian native build, with debian only dependencies, like molior aptly and debian official builds...

PR #1393, opened 2024-11-15
Improve/debianization

neolynx:
## Description of the Change

Synchronize with official debian packaging.


iofq:
Looks like the debug symbol file is potentially not working:

From CI build logs:
```
 dh_strip --dbg-package=aptly-dbg
dh_strip: warning: Could not find the BuildID in debian/aptly/usr/bin/aptly
readelf: Error: Unable to find program interpreter name
dh_strip: warning: Could not find the BuildID in debian/aptly-dbg/usr/lib/debug/usr/bin/aptly
omitted......
E: aptly-dbg: stripped-library [usr/lib/debug/usr/bin/aptly]
W: aptly-dbg: debug-file-with-no-debug-symbols [usr/lib/debug/usr/bin/aptly]
```

neolynx:
> Looks like the debug symbol file is potentially not working:

I think this is just a warning, and probably due to not building with debian dependencies. the aptly binary is stripped in the package, otherwise lintian would complain about this...

iofq (approved):

PR #1400, opened 2024-12-02
Fix: Null pointer when dropping a multi-dist published repo

cfiehe:
This commit fixes a null pointer when a multi-dist published repo gets dropped.

Signed-off-by: Christoph Fiehe <c.fiehe@eurodata.de>


neolynx (approved):
thanks !

PR #1401, opened 2024-12-02
Feature/yaml config

neolynx:
## Description of the Change

Documentation of the aptly configuration should be inline, which can be done with json:
- https://github.com/aptly-dev/aptly/blob/improve/documentation/debian/aptly.conf
However, this does not look that great.

In addition to json format config file, aptly can also support yaml format, which is easier to document:
- https://github.com/aptly-dev/aptly/blob/feature/yaml-config/debian/aptly.conf

This change brings the following:
- default configuration is documented yaml (created in home when no config file exists)
- `/etc/aptly.conf` in aptly-api debian package is documented yaml 
- zaml file contains default values, and example sections to be uncommented
- existing json configuration files are still valid
- json configuration can be converted to yaml with `aptly config show -yaml` (without documentation)

## Checklist
- [x] documentation updated


debian/aptly.conf
```diff
@@ -1,370 +1,339 @@
-// vim: : filetype=json
-// json configuration file with comments
-// validate with: sed '/\/\//d' aptly.conf | json_pp
-{
+# Aptly Configuration File
+###########################
 
-// Aptly Configuration File
-////////////////////////////
+# Aptly storage directory for:
```
neolynx, line 4: s/Aptly storage/Root/

iofq:
What is your reasoning for having new different keys in yaml for e.g rootDir vs root_dir? While i appreciate that yaml will be the default moving forward, having two different keys can cause confusion and I see little motivation for making a breaking change here

neolynx:
I was using the same syntax as in json, then I realized that snake case is recommended.
Since `aptly condif show -yaml` will convert existing json config, maybe it is not such a big migration (albeit the documentation comments will be missing).

we can still go with the json syntax if that is preferred.

neolynx:
Please vote for default (newly created) aptly config file to be:
- 👍: yaml with snake_case (root_dir, current)
- 🎉: yaml with same json files (rootDir)
- 🚀: json with documentation comments
- 😕: stay with current json config, undocumented

szakalboss:
snake case for consistency; btw snake type is python friendly ;)

szakalboss (approved):

PR #1411, opened 2024-12-20
Feature: Support Reading Filter Expressions from a File

### LINKED ISSUE #1409: Feature Suggestion: Support Reading Filter Expressions from a File
We have an internal patch for Aptly to support reading filter expressions from a file or from stdin. For example:

```shell
# from file
$ echo libc6 > test.filter; aptly mirror create -filter="@test.filter" ...
# from stdin
$ echo "vim" | aptly mirror create -filter='@-' ...
``` 

If there is any chance of getting this feature merged, I would clean up our code a bit and make a PR.

What do you think? 

## Detailed Description

For us it's mainly important that this works in the `mirror create` and `mirror edit` subcommands' `-filter` argument.
However, there are also other places in Aptly where a package query can be provided, so there is the question of whether we should support it everywhere for consistency, e.g.

```shell
aptly snapshot pull <name> <source> <destination> <package-query> ...
```

supports multiple queries, so should we allow the `@file` syntax there (then you could theoretically specify multiple different files and even mix it with plain queries like `@file1 vim @file2 ...`)?

Also, this is technically a breaking change, though I doubt many users have existing filters that start with `@` because that is not a valid character for a Debian package name and thus would not match any packages.

## Context

The reason we have implemented this is that we have filter expressions that exceed the command line length limit.

## Possible Implementation

Since the flag library used by Aptly (`smira/flag`) allows creating custom types of flags, I just made a `StringOrFileFlag` like this:

```go
// StringOrFileFlag is a custom flag type that can handle both string input and file input.
// If the input starts with '@', it is treated as a filename and the contents are read from the file.
// If the input is '@-', the contents are read from stdin.
type StringOrFileFlag struct {
	value string
}

func (s *StringOrFileFlag) Set(value string) error {
	var err error
	s.value, err = GetStringOrFileContent(value)
	return err
}

func GetStringOrFileContent(value string) (string, error) {
	if !strings.HasPrefix(value, "@") {
		return value, nil
	}

	filename := strings.TrimPrefix(value, "@")
	var data []byte
	var err error
	if filename == "-" { // Read from stdin
		data, err = io.ReadAll(os.Stdin)
	} else {
		data, err = os.ReadFile(filename)
	}
	if err != nil {
		return "", err
	}
	return string(data), nil
}

// ...
```

This way, the file/error handling logic is in one place and you can just use it like this without having to change any other code in other places:

```diff
- cmd.Flag.String("filter", "", "filter packages in mirror")
+ AddStringOrFileFlag(&cmd.Flag, "filter", "", "filter packages in mirror, use '@file' to read filter from file or '@-' for stdin")
```

## Your Environment

N/A

schoenherrg:
Fixes #1409

## Requirements

## Description of the Change

As mentioned in the Issue, we have filter expressions that exceed the command line length limit and thus need to be read from a file/stdin. This PR extends the Aptly command-line interface to be able to read filter expressions from a file using the `@/path/to/file` syntax or `@-` for stdin.

Our use case is only for the `-filter` argument for `mirror create`/`mirror edit`. However, there are other places in Aptly where filter expressions are accepted, so I added support for the `@file` syntax everywhere for consistency.

Commands where the `@file` syntax is supported now:

- `mirror create`
- `mirror edit`
- `package search`
- `package show`
- `repo move`
- `repo remove`
- `snapshot filter`
- `snapshot pull`
- `snapshot search`

## Checklist

- [X] documentation updated

### DIFF
--- a/cmd/mirror_create.go
+++ b/cmd/mirror_create.go
@@ -46,7 +46,7 @@ func aptlyMirrorCreate(cmd *commander.Command, args []string) error {
 		return fmt.Errorf("unable to create mirror: %s", err)
 	}
 
-	repo.Filter, err = getContent(context.Flags().Lookup("filter").Value.String())
+	repo.Filter = context.Flags().Lookup("filter").Value.String()
 	repo.FilterWithDeps = context.Flags().Lookup("filter-with-deps").Value.Get().(bool)
 	repo.SkipComponentCheck = context.Flags().Lookup("force-components").Value.Get().(bool)
 	repo.SkipArchitectureCheck = context.Flags().Lookup("force-architectures").Value.Get().(bool)
@@ -106,7 +106,7 @@ Example:
 	cmd.Flag.Bool("with-installer", false, "download additional not packaged installer files")
 	cmd.Flag.Bool("with-sources", false, "download source packages in addition to binary packages")
 	cmd.Flag.Bool("with-udebs", false, "download .udeb packages (Debian installer support)")
-	cmd.Flag.String("filter", "", "filter packages in mirror")
+	AddStringOrFileFlag(&cmd.Flag, "filter", "", "filter packages in mirror, use '@file' to read filter from file or '@-' for stdin")
 	cmd.Flag.Bool("filter-with-deps", false, "when filtering, include dependencies of matching packages as well")
 	cmd.Flag.Bool("force-components", false, "(only with component list) skip check that requested components are listed in Release file")
 	cmd.Flag.Bool("force-architectures", false, "(only with architecture list) skip check that requested architectures are listed in Release file")
--- a/cmd/mirror_create.go
+++ b/cmd/mirror_create.go
@@ -46,15 +46,12 @@ func aptlyMirrorCreate(cmd *commander.Command, args []string) error {
 		return fmt.Errorf("unable to create mirror: %s", err)
 	}
 
-	repo.Filter = context.Flags().Lookup("filter").Value.String()
+	repo.Filter = context.Flags().Lookup("filter").Value.String() // allows file/stdin with @
 	repo.FilterWithDeps = context.Flags().Lookup("filter-with-deps").Value.Get().(bool)
 	repo.SkipComponentCheck = context.Flags().Lookup("force-components").Value.Get().(bool)
 	repo.SkipArchitectureCheck = context.Flags().Lookup("force-architectures").Value.Get().(bool)
 
 	if repo.Filter != "" {
-		if err != nil {
-			return fmt.Errorf("unable to read package query from file %s: %w", context.Flags().Lookup("filter").Value.String(), err)
-		}
 		_, err = query.Parse(repo.Filter)
 		if err != nil {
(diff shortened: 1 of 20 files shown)


neolynx:
excellent feature :+1:

neolynx:
we should increase the coverage a bit...

 codecov/patch Failing after 1s — 60.49% of diff hit (target 75.09%)

neolynx:
@5hir0kur0 please push the branch to the main aptly-dev/aptly git, then the coverage runs automatically. otherwise I need to do a separate coverage pr.. :) could you open the PR there ?

cmd/string_or_file_flag.go
```diff
@@ -0,0 +1,55 @@
...
+func GetStringOrFileContent(value string) (string, error) {
+	if !strings.HasPrefix(value, "@") {
+		return value, nil
+	}
+
+	filename := strings.TrimPrefix(value, "@")
+	var data []byte
+	var err error
+	if filename == "-" { // Read from stdin
+		data, err = io.ReadAll(os.Stdin)
```
neolynx, line 47: shouldn't the CreateMirror37Test cover line 47 ?

cmd/string_or_file_flag.go
```diff
@@ -0,0 +1,55 @@
...
+	return s.value
+}
+
+func (s *StringOrFileFlag) Set(value string) error {
+	var err error
+	s.value, err = GetStringOrFileContent(value)
+	return err
+}
+
+func (s *StringOrFileFlag) Get() any {
```
neolynx, line 28: is this needed ?

cmd/mirror_create.go
```diff
@@ -52,6 +52,9 @@ func aptlyMirrorCreate(cmd *commander.Command, args []string) error {
 	repo.SkipArchitectureCheck = context.Flags().Lookup("force-architectures").Value.Get().(bool)
 
 	if repo.Filter != "" {
+		if err != nil {
```
neolynx, line 55: what `err` is checked here ?

cmd/string_or_file_flag.go
neolynx: yes, interface of Command

cmd/mirror_create.go
neolynx: seems useless code, removed

neolynx:
the coverage test does not seem to cover all of the code, when it is invoked by the flag parsing. 
I also tryed adding a test with a non existint @file, but was unable to get coverage output, as the program is aborted on error. 

let's accept low coverage here...

neolynx (approved):

cmd/mirror_create.go
schoenherrg: @neolynx Yes, thanks for removing. I forgot to remove that while refactoring the code.

schoenherrg:
@neolynx Thanks for your work!

> @5hir0kur0 please push the branch to the main aptly-dev/aptly git, then the coverage runs automatically. otherwise I need to do a separate coverage pr.. :) could you open the PR there?

Sorry, I didn't check my GitHub over the weekend, but I'll do it like that next time I make a PR!

PR #1423, opened 2025-01-11
Switch to google/uuid module

mikelolasagasti:
Current used github.com/pborman/uuid hasn't seen any updates in years.


neolynx:
thanks, that looks great !

could you also update debian/control to include the new build dependency ?

mikelolasagasti:
I think the new dep is already defined in https://github.com/aptly-dev/aptly/blob/master/debian/control#L27

neolynx:
> I think the new dep is already defined in https://github.com/aptly-dev/aptly/blob/master/debian/control#L27

oh, you are right !

then we only need to remove the golang-github-pborman-uuid-dev dependency. i will then test the pure debian build...

neolynx (approved):
thanks !

neolynx:
purde debian build works find without that dependency :+1:

PR #1440, opened 2025-04-14
Fix upload of unchanged packages in S3 on source update of published repository

### LINKED ISSUE #1435: publish switch to S3 uploading unchanged files starting in version 1.6
I am using `aptly` on macOS, installed via Homebrew. I have several S3 hosted repositories where I publish updates to like this:

```
aptly publish switch ... bookworm s3:${S3_BUCKET}: ${APTLY_SNAPSHOT_NAME}
```

Up through version 1.5.0 this worked as expected, and only new packages added to the repo were actually uploaded to S3, so the process was very quick.

Starting in version 1.6.1 (I didn't try in version 1.6.0 sorry) when I run the same `aptly publish switch` command, **all** files in the repository are uploaded to S3, even if they are not changed. I have let the command complete, and tried again several times thinking maybe the next times around it would not re-upload all files, but it still did.

I then reverted back to version 1.5.0, and `aptly publish switch` went back to working as expected: only newly added packages are uploaded to S3.

I noticed the 1.6 release notes mentioned changes to the S3 component of `aptly` so wondered if something broke with the `switch` command with those changes?

cfiehe:
Fixes #1435.

## Description of the Change

This PR fixes an issue in case of S3 storage that causes unchanged packages to be uploaded again on a source update of a published repository. The `storage.prefix` was added twice and the return paths from the storage did not contain the `pool` directory as first segment. That was the reason, why the code came to the conclusion, that the package does not exist in the bucket at the target location and uploaded the file again.


neolynx:
pipeline fixed. 
gocovmerge needed newer go verison, golangci-lint needed a newer version to be compatible ...

neolynx:
Articafts for testing can be found here: https://github.com/aptly-dev/aptly/actions/runs/14571537492

neolynx:
@cfiehe I wonder if we can somehow test this behavioe in a system test. maybe logging all uploads in debug mode, and the test would check the output ?

cfiehe:
> @cfiehe I wonder if we can somehow test this behavioe in a system test. maybe logging all uploads in debug mode, and the test would check the output ?

The question is: What should be covered by the test case? It would be the best, if you can somehow check, that only the expected files are uploaded to the S3 storage and validate afterwards that the files are present in the S3. I must admit, that I have not looked at the existing S3 test cases.

neolynx:
I managed to get this fix tested with the existing s3 tests, by adding
- initializing the logging for commands
- adding debug logs, which go to the system tests output

https://github.com/aptly-dev/aptly/actions/runs/14593800326/job/40934817222#step:10:754

cfiehe:
> I managed to get this fix tested with the existing s3 tests, by adding
> 
>     * initializing the logging for commands
> 
>     * adding debug logs, which go to the system tests output
> 
> 
> https://github.com/aptly-dev/aptly/actions/runs/14593800326/job/40934817222#step:10:754
> 
> ```
> +DBG S3: LinkFromPool 'pool/main/b/boost-defaults/libboost-program-options-dev_1.49.0.1_i386.deb'
> ```
> 
> With your fix reverted, the tests fail because the already uploaded package above is uploaded again. See: #1444
> 
> will cleanup and integrate here...

Ah, cool. That sounds good 👍.

neolynx (approved):

PR #1466, opened 2025-06-27
Fix non related API params from mirror update

### LINKED ISSUE #1463: mirror update via API issues (/api/mirrors/{name})
Endpoint: PUT /api/mirrors/{name}

Aptly version 1.6.2

I would imagine it runs an equivalent of: `aptly mirror update` , but it seems to do some very different logic, including creating a new mirror.

Why does it take as request body `api.mirrorUpdateParams`?

This endpoint ([/api/mirrors/{name}](https://github.dev/aptly-dev/aptly/blob/d1bfd29dfd8fa6f1e32a51c7e9249c2880a2bf52/api/mirror.go#L380)) seems to  create a new mirror based on `api.mirrorUpdateParams` [struct](https://github.dev/aptly-dev/aptly/blob/d1bfd29dfd8fa6f1e32a51c7e9249c2880a2bf52/api/mirror.go#L333) and even ignores the  "ArchiveRoot" of {name}, and update it while using the "Distribution" of the {name} mirror

Thanks for this project, very cool :)

itayporezky:
Fixes #1463 

I removed from the struct the parameters which are not part of [aptly mirror update](https://www.aptly.info/doc/aptly/mirror/update/)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -69,3 +69,4 @@ List of contributors, in chronological order:
 * Leigh London (https://github.com/leighlondon)
 * Gordian Schoenherr (https://github.com/schoenherrg)
 * Silke Hofstra (https://github.com/silkeh)
+* Itay Porezky (https://github.com/itayporezky)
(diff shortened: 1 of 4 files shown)


neolynx (approved):
thanks !

Issue #1475, opened 2025-08-01
S3 uploads all packages whether they are already present or not in published mirror


alguimodd:
Same issue as https://github.com/aptly-dev/aptly/issues/1435 that was fixed in https://github.com/aptly-dev/aptly/pull/1440

But I get the same problem. After looking at https://github.com/aptly-dev/aptly/pull/1440 I am not convinced that it fully solved the problem.

Version: v1.6.2 commit https://github.com/aptly-dev/aptly/commit/231039e8

Let's say we have a published mirror with prefix `test/ubuntu-ports-small` that includes package `nginx` and `nginx-core` from `ubuntu-ports-main`.
I included a debug line
```
s3/public.go
		for i := range paths {
			storage.pathCache[filepath.Join("pool", paths[i])] = md5s[i]
		}
		fmt.Printf("o=LinkFromPool pathCache: %+v\n", storage.pathCache)
```
First time we publish we get the following output:
```
./build/aptly publish snapshot \
              -distribution=jammy-small \
              -architectures=arm64 \
              -skip-signing \
              -component=main \
              -skip-contents \
              jammy-main-arm64-small \
              "s3:aptly-s3:test/ubuntu-ports-small"
Loading packages...
Generating metadata files and linking package files...
 0 / 2 [------------------------------------------------------------------------------------------------------------------------------------------------]   0.00
%o=LinkFromPool pathCache: map[]
DBG S3: LinkFromPool 'test/ubuntu-ports-small/pool/main/n/nginx/nginx_1.18.0-6ubuntu14_arm64.deb'
 1 / 2 [=======================================================================>------------------------------------------------------------------------]  50.00%
DBG S3: LinkFromPool 'test/ubuntu-ports-small/pool/main/n/nginx/nginx-core_1.18.0-6ubuntu14_arm64.deb'
Finalizing metadata files...
```
Which indeed indicates that the pool was empty `map[]`.
If we switch to the same snapshot, the packages do not need to be published again, and that is what https://github.com/aptly-dev/aptly/pull/1440 solved. But in my case I get:
```
./build/aptly publish switch \
              -architectures=arm64 \
              -skip-signing \
              -component=main \
              -skip-contents \
              jammy-small \
              "s3:aptly-s3:test/ubuntu-ports-small" \
              jammy-main-arm64-small
Loading packages...
Generating metadata files and linking package files...
 0 / 2 [------------------------------------------------------------------------------------------------------------------------------------------------]   0.00%
[...]

PR #1480, opened 2025-08-13
feat(s3): add publishedPrefix to pathCache to avoid reupload of files

alguimodd:
Fixes https://github.com/aptly-dev/aptly/issues/1475

## Requirements

## Description of the Change

This adds the publishedPrefix to paths added to `pathCache` so it can be correctly compared to the `relPath`.

## Checklist

- [ ] documentation updated

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,10 +109,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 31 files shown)


cfiehe (approved):
This change should fix the cache lookup issue. Files are looked up with respect to the `publishedPrefix`, but they are not put into the cache containing that `publishedPrefix`. The result is a cache miss in use cases with a non-empty `publishedPrefix`.

@alguimodd
Thanks yor fixing and your detailed analysis.

@neolynx
Can you please merge this change and close the corresponding issue. Thanks a lot.

PR #1484, opened 2025-08-21
Support updating label and origin domain of publish

### LINKED ISSUE #1007: Add option to change settings of publish without dropping?
Hi,

Currently when you need to change certain parameters of a published repo (from snapshot) then you need to drop the entire published repo (meaning during that time it won't be available) and then republish it with the new data.

For example if you need to just enable support for acquire-by-hash you cannot do it via switch. Or change Origin/Label etc.

Is there any technical limitation behind this? 

Thanks

xzhang1:
Fixes #1007

## Requirements

## Description of the Change

Some times when making the publish, "Label" and "Origin" are not set, or not set correctly thus need to be modified later. With this modification, "Label" and "Origin" can be modified through publish update, no need to drop the re-build the whole publish.

## Checklist

- [x] documentation updated


xzhang1:
Support update "label" and "origin" of aptly publish on:

1) Command `aptly publish update`
Add new test case "t06_publish/PublishUpdate19Test"

2) REST API: `api.POST("/publish/:prefix/:distribution/update", apiPublishUpdate)`
Modify test case "t12_api/PublishUpdateSourcesAPITestRepo"

3) REST API: `api.PUT("/publish/:prefix/:distribution", apiPublishUpdateSwitch)`
Modify test case "t12_api/PublishSwitchAPITestRepo"

neolynx:
I renamed on my branch already... but I think there is a problem in the test itself.. investigating...

Issue #1474, opened 2025-07-25
Uploads return success when server is out of space


russelltg:
We have a server that is out of space, but all the uploads/publishes were marked as successful. 

It seems we should get API errors when the OS returns ENOSPC?

PR #1504, opened 2025-11-13
Return error on out of space

bwitt:
Fixes https://github.com/aptly-dev/aptly/issues/1474

## Requirements

## Description of the Change

Return an error if we run out of disk space.  Batch up the sync at the end of writing all files.

## Checklist

- [ ] documentation updated

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ env:
 
 jobs:
   unit-test:
-    name: "Unit Tests (Debian 13)"
+    name: "Unit Tests"
     runs-on: ubuntu-22.04
     continue-on-error: false
     timeout-minutes: 30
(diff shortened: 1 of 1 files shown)


neolynx:
looks good !

could you add yourself to the AUTHORS file ?

bwitt:
Done!  ls the added coverage enough?

bwitt:
maybe we can try running these tests with sudo

bwitt:
> > ```
> > [2026-01-05T00:30:21.971Z] ['error'] There was an error running the uploader: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 429 - {"message":"Rate limit reached. Please upload with the Codecov repository upload token to resolve issue. Expected time to availability: 482s."}
> > ```

Actually I think this missing codecov token is expected. this is running on my fork, which does not have access to the secret `CODECOV_TOKEN` which is by design for forks.

neolynx:
codecov/patch — 75.55% of diff hit (target 74.90%)
codecov/project — 75.41% (+0.50%) compared to a70f572

way better... ;-) let me update the PR

neolynx:
I added 3 changes:

- ci: run unit tests in docker
- ci: provide 1MB /smallfs to docker
- unit-test: use /smallfs when non-root

now, when not running as root, a /smallfs mount is expected. Pipeline and make docker-unit tests start a docker container providing a 1MB /smallfs...

maybe the coverage upload should fail the pipeline if it cannot upload ?

bwitt:
thanks!  I see the codecov run failed, and it failed CI.
per https://docs.codecov.com/docs/codecov-tokens#uploading-without-a-token

> Please update to the latest versions of our CI actions or the Codecov CLI for more reliable tokenless uploads.

we could try using the latest codecov action v5 and see if that helps.

(separately: do we not have an automatic dependency bumper like Dependabot configured for this repo, so that this would have been updated already?)

Also sounds like uploading without a token is expected to fail, so I'm not sure if we want this to fail CI on PRs from forks like this, since it will fail, and it's expected to fail.

PR #1508, opened 2025-11-27
fixed urlencode for repo name in /put

### LINKED ISSUE #1507: PUT /api/repos/<name> doesn't work with non-url characters
when repo has a name such as `repo@test` aptly fails with PUT. no matter will @ be urlencoded or not inside path, method only works if inside json `Name` will be encoded as well

PUT api/repos/repo@test --data '{"Comment": "example repo", "Name": "repo@test"}' -> ❌
PUT api/repos/repo%40test --data '{"Comment": "example repo", "Name": "repo@test"}' -> ❌
PUT api/repos/repo@test --data '{"Comment": "example repo", "Name": "repo%40test"}' -> ❌

PUT api/repos/repo%40test --data '{"Comment": "example repo", "Name": "repo%40test"}' -> works

which is inconsistent because any other method (GET, POST, DELETE): 1) receives @ in url 2) prints repo info with @

theoborealis:
Fixes #1507

## Description of the Change

Changes wrong comparison taking url encode into account

## Checklist

### DIFF
--- a/api/repos.go
+++ b/api/repos.go
@@ -167,7 +167,7 @@ func apiReposCreate(c *gin.Context) {
 
 type reposEditParams struct {
 	// Name of repository to modify
-	Name *string `binding:"required"     json:"Name"                 example:"repo1"`
+	Name *string `                       json:"Name"                 example:"new-repo-name"`
 	// Change Comment of repository
 	Comment *string `                    json:"Comment"              example:"example repo"`
 	// Change Default Distribution for publishing
@@ -179,7 +179,7 @@ type reposEditParams struct {
 // @Summary Update Repository
 // @Description **Update local repository meta information**
 // @Tags Repos
-// @Param name path string true "Repository name"
+// @Param name path string true "Repository name to modify"
 // @Consume  json
 // @Param request body reposEditParams true "Parameters"
 // @Produce json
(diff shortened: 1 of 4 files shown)


neolynx:
Would this also apply to all other methods receiving a 'name' param ?

neolynx (approved):

neolynx:
@theoborealis could you wanna add your self to the AUTHORS file ?

also we should add a little system test to keep the coverage.

you could also allow edit access to the maintainers for this PR, then I can help with the testing...

neolynx:
Hi @theoborealis,

I have added a test for repoEdit, coverage is up a bit :)

then I looked a bit further into the change, and think the name field from the URL is already escaped, like all the other url parameters in other functions. what was not escaped, is the optional new name to set was not escaped. 

I think the current changes solve the problem, could yu give it a try and confirm ?

neolynx:
Hi @theoborealis,

I am still not quite clear what we are fixing here.. 
-> the repo param in the url is properly unesaped (and escaped by the http client)
-> the json body does not need to be escaped, normal strings can be passed

could you provide an example of what is not working in details ? 
the calls provided https://github.com/aptly-dev/aptly/issues/1507 seem to work for me, even without your change...

neolynx:
I believe this is fixed / not an issue anymore

Issue #1497, opened 2025-10-07
[FR?] `InRelease` `Signed-By` field support


LebedevRI:
https://wiki.debian.org/DebianRepository/Format#Signed-By says:
> **Signed-By**
> An optional field containing a comma separated list of OpenPGP key fingerprints to be used for validating the next Release file. The fingerprints must consist only of hex digits and may not contain spaces. The fingerprint specifies either the key the Release file must be signed with or the key the signature key must be a subkey of. The later match can be disabled by appending an exclamation mark to the fingerprint.
>
> If the field is present, a client should only accept future updates to the repository that are signed with keys listed in the field. The field should be ignored if the Valid-Until field is not present or if it is expired.

Unless i'm really missing it, aptly does not produce this field,
and can not be configured to produce this field. Is that correct?

It would be nice if the value for that field could be configured,
with MVP being treating it as a opaque textual field without actual validation,
that can be configured during `aptly publish (snapshot|repo)`.

Thank you!

PR #1518, opened 2025-12-30
`InRelease` file: support `Signed-By` field

LebedevRI:
https://wiki.debian.org/DebianRepository/Format#Signed-By says:
> **Signed-By**
> An optional field containing a comma separated list of
> OpenPGP key fingerprints to be used for validating
> the next Release file. The fingerprints must consist
> only of hex digits and may not contain spaces.
> The fingerprint specifies either the key the Release file
> must be signed with or the key the signature key must be
> a subkey of. The later match can be disabled by appending
> an exclamation mark to the fingerprint.
>
> If the field is present, a client should only accept future updates
> to the repository that are signed with keys listed in the field.
> The field should be ignored if the Valid-Until field is not present
> or if it is expired.

For both the CLI tools and JSON, the field is taken as a string verbatim.

When specified, we must also provide `Valid-Until` field, and i'm not sure there is an 'infinity' value for it, so 100 years will have to do?

Fixes https://github.com/aptly-dev/aptly/issues/1497

## Description of the Change

This functionality is missing as compared to reprepro,
and i would have preferred not to have lost it when switching `apt.bcachefs.org` repo.
It's not terribly important, but hey, it's not like it's too intrusive to the users.

## Checklist

- [ ] documentation updated


neolynx (approved):
awesome, thanks !

PR #1524, opened 2026-01-07
Update Puppet module references in README

jay7x:
Removed outdated Puppet module references and added an actively maintained one.


neolynx (approved):
Thanks !

PR #1537, opened 2026-02-20
Add SOURCE_DATE_EPOCH support for reproducible builds

tonobo:
## Description of the Change

Reproducible builds ensure that building the same source input always produces byte-identical output, regardless of when or where the build happens. This is important for security and trust: anyone can independently verify that published binaries actually correspond to their source code, with no tampering or hidden modifications.

When aptly publishes a repository to the filesystem, the Release file includes a Date field set to the current time. This means publishing the same packages twice produces different output, breaking reproducibility. By honoring SOURCE_DATE_EPOCH (a Unix timestamp typically derived from the last source change), the Date and Valid-Until fields become deterministic, making the entire published repository output byte-identical across runs with the same input.
This follows the https://reproducible-builds.org/specs/source-date-epoch/ specification, which is widely adopted across the Debian ecosystem and other projects.

## Checklist

- [ ] documentation updated


tonobo:
@neolynx Can you please have a look?

neolynx (approved):
awesome job, thanks !

tonobo:
Thanks man!

PR #1542, opened 2026-03-03
Add SOURCE_DATE_EPOCH support for GPG signers

tonobo:
Both the external GPG signer (`--faked-system-time`) and internal Go OpenPGP signer (`signerConfig.Time`) now honor SOURCE_DATE_EPOCH, producing reproducible signatures alongside the plain Release file dates.

Adds system tests for both signer backends verifying byte-identical Release, Release.gpg and InRelease across repeated publishes.

The signer tests (PublishRepo3[78]Test) are using an ed25519 key because
ed25519 signatures are deterministic by design. The Go openpgp library
uses a random nonce for DSA/ECDSA (see signature.go Sign calls using
config.Random() link below) so those signatures vary across runs
even with a fixed timestamp, making byte-identical verification impossible.

In addition to 49f342878ad2b220d2f870dc69f71ec05e26f42e 

Ref: https://github.com/aptly-dev/aptly/pull/1537
Ref: https://github.com/ProtonMail/go-crypto/blob/v1.4.0/openpgp/packet/signature.go#L945-L979

## Checklist

- [x}  functional test added/updated (if change is functional)
- [ ] documentation updated


tonobo:
@neolynx sorry for the inconvenience, I had to submit a follow-up. The original commit only works as long as the repository is not signed. I'm personally using the internal signer only, which causes issues on two fronts.

The fix itself is fairly simple, as you may have seen in the changes. However, aptly uses an outdated version of the PGP package, which isn't a problem on its own. The complication is that I'm importing aptly not as a prebuilt binary but as a cmd wrapper. There are various reasons for this (admittedly ugly) approach, but the downside / or upside, depending on your perspective, is that we needed to align the go-crypto version with at least v1.4, where `NonDeterministicSignaturesViaNotation` was introduced.

Always fun :smile:

neolynx:
rebased on master,,,

had to rename tests.

neolynx (approved):
thanks !

Issue #1467, opened 2025-07-04
Fatal error when publishing more than 1 repository to S3 concurrently


michel-ea:
Hi,

Using the latest version (1.6.2) we get a fatal go exception when publishing multiple repositories to S3 concurrently. This is using the REST API `/publish/s3:S3:<repo>/<distro>` endpoint. 

Exception:

```
2025-06-24 14:53:13.754	fatal error: concurrent map writes
2025-06-24 14:53:13.757	
2025-06-24 14:53:13.757	goroutine 4073 [running]:
2025-06-24 14:53:13.760	github.com/aptly-dev/aptly/s3.(*PublishedStorage).LinkFromPool(0xc003612f80, {0xc003626108, 0x18}, {0xc008df0b80?, 0x0?}, {0xc003877ad0, 0x23}, {0x55b37ea1da60, 0xc00265e6a0}, {0xc002464000, ...}, ...)
2025-06-24 14:53:13.760		/__w/aptly/aptly/s3/public.go:390 +0x88d
2025-06-24 14:53:13.760	github.com/aptly-dev/aptly/deb.(*Package).LinkFromPool(0xc0016e93f0, {0x55b37ea1f9a0, 0xc003612f80}, {0x55b37ea1da60, 0xc00265e6a0}, {0xc003626108, 0x18}, {0xc008df0b80, 0x1b}, 0x0)
2025-06-24 14:53:13.760		/__w/aptly/aptly/deb/package.go:698 +0x1f6
2025-06-24 14:53:13.760	github.com/aptly-dev/aptly/deb.(*PublishedRepo).Publish.func2(0xc0016e93f0)
2025-06-24 14:53:13.760		/__w/aptly/aptly/deb/publish.go:942 +0x745
2025-06-24 14:53:13.760	github.com/aptly-dev/aptly/deb.(*PackageList).ForEachIndexed(0xc000ef1400?, 0xc006719c10)
2025-06-24 14:53:13.760		/__w/aptly/aptly/deb/list.go:183 +0x70
2025-06-24 14:53:13.760	github.com/aptly-dev/aptly/deb.(*PublishedRepo).Publish(0xc00003e240, {0x55b37ea1da60, 0xc00265e6a0}, {0x55b37ea00f48?, 0xc000d7c000?}, 0xc000180320, {0x0, 0x0}, {0x55b37ea1fb60, 0xc003e40860}, ...)
[...]

Issue #1449, opened 2025-05-02
packages missing after concurrent publish


flixr:
We are still seeing the issue #1125 with concurrent publish once in a while with 1.6.0:

E.g. after uploading (to different tmp folders) and publishing two packages in parallel:
```
[GIN] 2025/04/09 - 16:55:59 | 200 |    8.029972ms |       10.0.0.42 | POST     "/api/files/tmp.cAIGfOwQt99ukdk"
[GIN] 2025/04/09 - 16:55:59 | 200 |    7.242806ms |       10.0.0.45 | POST     "/api/files/tmp.zO8MZsfCgL7u2Uw"
[GIN] 2025/04/09 - 16:55:59 | 200 |   94.609874ms |       10.0.0.45 | POST     "/api/repos/roboception-jammy-testing/file/tmp.zO8MZsfCgL7u2Uw/foobar_0.21.0+11+g1f08260-0+20250409+164139~jammy_arm64.deb"
[GIN] 2025/04/09 - 16:55:59 | 200 |  228.518648ms |       10.0.0.42 | POST     "/api/repos/roboception-focal-testing/file/tmp.cAIGfOwQt99ukdk/foobar_0.21.0+11+g1f08260-0+20250409+164136~focal_arm64.deb"
Signing file 'Release' with gpg, please enter your passphrase when prompted:
Clearsigning file 'Release' with gpg, please enter your passphrase when prompted:
[GIN] 2025/04/09 - 16:56:05 | 200 |  5.673881188s |       10.0.0.45 | PUT      "/api/publish/testing/jammy"
Signing file 'Release' with gpg, please enter your passphrase when prompted:
Clearsigning file 'Release' with gpg, please enter your passphrase when prompted:
[GIN] 2025/04/09 - 16:56:09 | 200 |  9.535539358s |       10.0.0.42 | PUT      "/api/publish/testing/focal"
```

Afterwards `foobar_0.21.0+11+g1f08260-0+20250409+164136~focal_arm64.deb` is listed as a package, but trying to download it fails:
```
E: Failed to fetch http://our-apt-server.com/testing/pool/main/f/foobar/foobar_0.21.0+11+g1f08260-0+20250409+164136~focal_arm64.deb  404  Not Found [IP: 10.0.0.1 80]
```
The file is actually there in
`./pool/e4/be/af861880689cd3d9b8384504f350_foobar_0.21.0+11+g1f08260-0+20250409+164136~focal_arm64.deb`
but the file
[...]

Issue #1376, opened 2024-10-21
Lost packages after publication


ishaulov:
Hi! After resolved https://github.com/aptly-dev/aptly/issues/1276 I builded and installed `aptly `version `1.5.0+207+g372ce3c4`

And now some big packages (>900 Mb) are lost with concurent builds on Jenkins.
When uploading and publishing the package I get status 2, but the package is not in the repository.
For example, the package `testfile_4.4.5.7667` was lost.

Please, help me understand, why this is happening?

```
root@testrepo:~# aptly repo show -with-packages testrepo
testfile_4.4.5.7665_amd64
testfile-archive-_4.4.5.7667_amd64
testfile-long-data_4.4.5.7667_amd64
```

In aptly access.log file:
```
"level":"info","remote":"192.168.100.113","method":"POST","path":"/api/repos/testrepo/file/tmp.U5FXf4EUJcvNTWu/testfile_4.4.5.7667_amd64.deb?_async=true","protocol":"HTTP/1.0","code":"202","latency":"231.621µs","agent":"curl/7.58.0","time":"2024-10-20T22:56:47+05:00"}
```

In Jenkins console output:
```
23:08:36 + publish_output='+++ mktemp -d tmp.XXXXXXXXXXXXXXX
23:08:36 ++ upload_folder=tmp.U5FXf4EUJcvNTWu
23:08:36 ++ curl -f -X POST -F file=@/home/lnxbuilder/7667/deb//testfile_4.4.5.7667_amd64.deb http://localrepo.com/deb/api/files/tmp.U5FXf4EUJcvNTWu -u login:pass
23:08:36 ++ json_pp
23:08:36   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
23:08:36                                  Dload  Upload   Total   Spent    Left  Speed
23:08:36 
23:08:36   0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
23:08:36   0 1738M    0     0    0  480k      0  18.0M  0:01:36 --:--:--  0:01:36 18.0M
23:08:36   2 1738M    0     0    2 51.0M      0  49.7M  0:00:34  0:00:01  0:00:33 49.7M
23:08:36   5 1738M    0     0    5  101M      0  50.1M  0:00:34  0:00:02  0:00:32 50.1M
23:08:36  10 1738M    0     0   10  182M      0  60.1M  0:00:28  0:00:03  0:00:25 60.1M
[...]

Issue #1196, opened 2023-07-05
Concurrent map writes crash after publish


russelltg:
## Detailed Description

Aptly crashed with a concurrent map writes error. Log: https://gist.github.com/russelltg/6ece2a67e723b2fb7560851a69257055

This has happened a few times now, seems to happen after every few weeks or so. Let me know if there's any additional info I can provide

## Your Environment

```bash
$ aptly version
aptly version: '1.5.0+48+gecc05518'
```

neolynx:
I think this is fixed on master, as many race confidions have been eliminated. could you please verify and close this issue ?

russelltg:
I haven't seen this in a while. Let's close for now and I'll open a new report if I see something new.

PR #1574, opened 2026-05-25
Fix concurrent API race conditions

### LINKED ISSUE #1570: getFileLock in files/public.go leaks memory via unbounded fileLocks map
The LinkFromPool concurrency fix (commit 3608c137 / PR #1449) introduced a per-path mutex map to prevent race conditions during concurrent publish operations:
```
    var (
        fileLockMutex sync.Mutex
        fileLocks = make(map[string]*sync.Mutex)
    )

    func getFileLock(filePath string) *sync.Mutex {
        fileLockMutex.Lock()
        defer fileLockMutex.Unlock()

        if mutex, exists := fileLocks[filePath]; exists {
            return mutex
        }
        mutex := &sync.Mutex{}
        fileLocks[filePath] = mutex
        return mutex
    }
```
The problem is that fileLocks is append-only. Every unique destinationPath ever passed to LinkFromPool allocates a *sync.Mutex entry that is never removed. This is a slow but unbounded memory leak - estimated impact is around 150-200 bytes per unique path, so after 100,000 packages are registered I'd expect around ~20MB of extra memory used. 

That's small enough that it might be worth just accepting as a tradeoff as attempting to prevent this definitely adds a lot of complexity - either adding a refcount to fileLocks, or storing these in a Least Recently Used list where we can bound the size of filelocks and once we hit that capacty we start evicting the oldest ones.

neolynx:
Fixes #1570, #1467, #1449, #1376, #1305, #1196, #1163, #1125

## Description of the Change

This fixes several race conditions when running API calls concurrently.

 * lock snapshots/repos on publish
 * lock pool on MultiDist=false
 * use UUID instead Name for locking, since rename operations might leave locks on outdated names
 * perform collection.LoadComplete inside maybeRunTaskInBackground
 * have tasks use a fresh copy of taskCollectionFactory, taskCollection
 * fix toggling between MultiDist=false/true

## Checklist


cchazalet:
Hi @neolynx,

I just tested the switch MultiDist option (True->False and False->True) in Jenny, and I can confirm that it works great! 

No need to run “publish drop” and “publish create” anymore, thanks !

cchazalet (approved):
I'm observing many concurrent publications when MultiDist=True and publications one after another when MultiDist=False

No race conditions observed

PR #1598, opened 2026-06-29
Dont lock the go version to the .0 release

C0deSentin3l:
the .0 release of go 1.25 is not the latest version of that release so missing out on security fixes etc

Ideally aptly should be on the 1.26 release as that is the latest stable version as when 1.27 of golang is released 1.25 will no longer be maintained, if you need a easy way to keep golang up to date on you system take a look at  
which is my personal github account https://github.com/L1ghtn1ng/go-updater

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/aptly-dev/aptly
 
-go 1.25.0
+go 1.25
 
 require (
 	github.com/AlekSi/pointer v1.1.0


neolynx (approved):
thanks!

neolynx:
hmm, the problem is, now the pipeline fails: https://github.com/aptly-dev/aptly/actions/runs/28381445064/job/84084829195#step:8:58

when I run go mod tidy, it puts 1.25.0 in the go.mod again... 

@C0deSentin3l and ideas about this ?

neolynx:
I moved to go 1.26, since this allows to put 1.26 and go mod tidy does not complain. also, 1.26 is available in debian, and aptly aims to be compatible. so using latest go always is not a path for us...

C0deSentin3l:
@neolynx sorry did not see your message till now

C0deSentin3l:
@neolynx I suspect its down to your github actions being very out of date, I recommend you pull this in to update your github actions and also your deps as well https://github.com/L1ghtn1ng/traceguard/blob/main/.github/dependabot.yml also adding this would be a good idea to the build https://github.com/L1ghtn1ng/traceguard/blob/17127e3b6ba3440494497e7197b6b24fcf759613/Makefile#L7 will increase the hardening of the binary

C0deSentin3l:
I also recommend you get this installed https://github.com/zizmorcore/zizmor and then do zizmor .github/ --fix=all this will also help sure up the ci as well

PR #1601, opened 2026-07-06
Fix version parsing, validate debian package version

jaymzh:
This patch is based on one originally by Aaron Foster
<afoster@cloudflare.com>, to prevent invalid versions.

However, applying it revealed an underlying issue in the Aptly version
parsing - the spec says the "debian version" is the part after the
_last_ hyphen, not the first hyphen, and that caused packages that were
valid to fail validation, so I've fixed the parsing here, and updated
the tests.

Signed off with internal and external emails for clarity, but commit
is under personal email where we typically do open source contributions
from.

Signed-off-by: Phil Dibowitz <pdibowitz@cloudflare.com>
Signed-off-by: Phil Dibowitz <pdibowitz@ipom.com>

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -84,3 +84,4 @@ List of contributors, in chronological order:
 * Zhang Xiao (https://github.com/xzhang1)
 * Tom Nguyen (https://github.com/lecafard)
 * Philip Cramer (https://github.com/PhilipCramer)
+* Phil Dibowitz (https://github.com/jaymzh)
--- a/deb/import.go
+++ b/deb/import.go
@@ -124,6 +124,12 @@ func ImportPackageFiles(list *PackageList, packageFiles []string, forceReplace b
 			continue
 		}
 
+		if !isValidVersion(p.Version) {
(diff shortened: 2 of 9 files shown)


r4co0n:
When you are talking about "debian versions", you mean Debian revisions (`debian_revision`) as specified in https://www.debian.org/doc/debian-policy/ch-controlfields.html#version , right? It is correct that the last hyphen(`-`) character marks the beginning of the Debian revision.

EDIT: Either there is no Debian revision in the version string, then hyphens are not allowed in `upstream_version`, or the last hyphen marks the Debian revision.

jaymzh:
@r4co0n yes, that's correct. You are saying the same thing I am.

jaymzh:
Debian tests are failing due to upstream repos for bullseye:
```
E: Release file for http://deb.debian.org/debian-security/dists/bullseye-security/InRelease is expired (invalid since 3d 20h 9min 32s). Updates for this repository will not be applied.
```

PR #1603, opened 2026-07-07
Expose net/http/pprof handlers

jmunson:
Expose net/http/pprof handlers under /api/debug/pprof/* when the new enablePprofEndpoint (JSON) / enable_pprof_endpoint (YAML) config option is true. The endpoint is disabled by default.

These are useful for adhoc debugging orperformance analysis, or ongoing performance monitoring when polled with Grafana Alloy

Tests have been added to demonstrate the endpoints responding when enabling and not when not enabled - I'm unsure how much value that is adding but wanted to err on the safe side of test coverage, but feel free to drop them if they are superfluous. 
## Checklist

- [ ] documentation updated

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -84,3 +84,4 @@ List of contributors, in chronological order:
 * Zhang Xiao (https://github.com/xzhang1)
 * Tom Nguyen (https://github.com/lecafard)
 * Philip Cramer (https://github.com/PhilipCramer)
+* James Munson (https://github.com/jmunson)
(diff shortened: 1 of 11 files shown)


jmunson:
Sorry when I wrote this I only ran my own new tests and didn't notice I broke the other tests, so I'm chipping away at those still

jmunson:
I believe the above timeout is unrelated so I'll leave this as is but let me know if you'd like me to take a deeper look

(edit: triggered it again with a no-op change, seems fine now)

PR #1606, opened 2026-07-10
Split reflists to share their contents across snapshots [take 2]

refi64:
## Description of the Change

(See the bottom of this message for the changes since #1235.)

In current aptly, each repository and snapshot has its own reflist in the database. This brings a few problems with it:

- Given a sufficiently large repositories and snapshots, these lists can get enormous, reaching >1MB. This is a problem for LevelDB's overall performance, as it tends to prefer values around the confiruged block size (defaults to just 4KiB).
- When you take these large repositories and snapshot them, you have a full, new copy of the reflist, even if only a few packages changed. This means that having a lot of snapshots with a few changes causes the database to basically be full of largely duplicate reflists.
- All the duplication also means that many of the same refs are being loaded repeatedly, which can cause some slowdown but, more notably, eats up huge amounts of memory.
- Adding on more and more new repositories and snapshots will cause the time and memory spent on things like cleanup and publishing to grow roughly linearly.

*Split reflists* aim at solving this by separating reflists into 64 *buckets*. Package refs are sorted into individual buckets according to the following system:

- Take the first 3 letters of the package name, after dropping a `lib` prefix. (Using only the first 3 letters will cause packages with similar prefixes to end up in the same bucket, under the assumption that packages with similar names tend to be updated together.)
- Take the 64-bit xxhash of these letters. (xxhash was chosen because it relatively good distribution across the individual bits, which is important for the next step.)
- Use the first 6 bits of the hash (range [0:63]) as an index into the buckets.

Once refs are placed in buckets, a sha256 digest of all the refs in the bucket is taken. These buckets are then stored in the database, split into roughly block-sized segments, and all the repositories and snapshots simply store an array of bucket digests.

This approach means that *repositories and snapshots can share their reflist buckets*. If a snapshot is taken of a repository, it will have the same contents, so its split reflist will point to the same buckets as the base repository, and only one copy of each bucket is stored in the database. When some packages in the repository change, only the buckets containing those packages will be modified; all the other buckets will remain unchanged, and thus their contents will still be shared. Later on, when these reflists are loaded, each bucket is only loaded once, short-cutting loaded many megabytes of data. In effect, split reflists are essentially copy-on-write, with only the changed buckets stored individually.

Changing the disk format means that a migration needs to take place, so that task is moved into the database cleanup step, which will migrate reflists over to split reflists, as well as delete any unused reflist buckets.

All the reflist tests are also changed to additionally test out split reflists; although the internal logic is all shared (since buckets are, themselves, just normal reflists), some special additions are needed to have native versions of the various reflist helper methods.

- Memory usage during publish and database cleanup, with `GOMEMLIMIT=2GiB`, goes down from ~3.2GiB (larger than the memory limit!) to ~0.7GiB, a decrease of ~4.5x.

*In my local tests*, publish times had also decreased down to mere seconds but the same effect wasn't observed on the server, with the times staying around the same. My suspicions are that this is due to I/O performance: my local system is an M1 MBP, which almost certainly has much faster disk speeds than our DigitalOcean block volumes. Split reflists include a side effect of requiring more random accesses from reading all the buckets by their keys, so if your random I/O performance is slower, it might cancel out the benefits. That being said, even in that case, the memory usage and database size advantages still persist.

---

Since #1235 and #1282, there were a few improvements:

- I did a fresh rebase on the latest git branch.
- Some strange / problematic edge-case behavior and error handling was fixed.
- Tests were improved.

During this time, we've kept split reflists running on our infra, with no major issues stemming from it. Additionally, the number of snapshots in the database grew to **over 200k**, with nearly 70k of those actually published (the reason for this difference is some gaps in our cleanup, oops). Despite that growth, the total database size is ~1.5GiB, which is barely larger than our original database was with just a few thousand snapshots.

## Checklist

- [ ] documentation updated


neolynx:
thanks a lot for reviving this !

shall we close https://github.com/aptly-dev/aptly/pull/1282 ?

refi64:
Yeah I think we can close it (originally I was going to try and reuse the extra fix commits there, but rebasing multiple commits is more annoying, and there were new lints to fix anyway, so it was easier to just start from scratch).

refi64:
Just squashed in a very small additional change to clean up a place in `deb/publish.go` where I was still using regular reflists for some reason

```
diff --git a/deb/publish.go b/deb/publish.go
index 9ceb449a7f..100dfe9faf 100644
--- a/deb/publish.go
+++ b/deb/publish.go
@@ -1542,7 +1542,7 @@
 func (collection *PublishedRepoCollection) listReferencedFilesByComponent(prefix string, components []string,
 	collectionFactory *CollectionFactory, progress aptly.Progress) (map[string][]string, error) {
 	referencedFiles := map[string][]string{}
-	processedComponentRefs := map[string]*PackageRefList{}
+	processedComponentRefs := map[string]*SplitRefList{}
 
 	processedComponentBuckets := map[string]*RefListDigestSet{}
 	for _, component := range components {
@@ -1580,12 +1580,12 @@
 						}
 						processedBuckets.Add(digest)
 
-						unseenRefs := bucket
+						unseenRefs := NewSplitRefListFromRefList(bucket)
 						processedRefs := processedComponentRefs[component]
 						if processedRefs != nil {
 							unseenRefs = unseenRefs.Subtract(processedRefs)
 						} else {
-							processedRefs = NewPackageRefList()
+							processedRefs = NewSplitRefList()
 						}
 
 						if unseenRefs.Len() == 0 {
```

This can result in ~50% perf savings when processing large numbers of snapshots. Somehow I missed it first time around.

refi64:
One more tiny change I pushed up: I was mistakenly taking the reflist digest every time a bucket was used, rather than just when they're loaded, so it was running sha256 relative to the number of reflists (bad) rather than the number of buckets (good, this is the whole point of split reflists!).

(Interestingly, I never noticed it as a hotspot before, because of a mix of sha256 still being pretty fast + my local system has hardware accelerated sha256 support.)

```diff
diff --git a/deb/reflist.go b/deb/reflist.go
index 262d118c77..10066c6666 100644
--- a/deb/reflist.go
+++ b/deb/reflist.go
@@ -897,17 +897,17 @@
 				// The segments may not have been iterated in order, so make sure to re-sort
 				// here.
 				sort.Sort(bucket)
+
+				actualDigest := reflistDigest(bucket)
+				if !bytes.Equal(actualDigest, digest) {
+					return fmt.Errorf("corrupt reflist bucket %d: expected digest %s, got %s",
+						idx,
+						bucketDigestEncoding.EncodeToString(digest),
+						bucketDigestEncoding.EncodeToString(actualDigest))
+				}
+
 				collection.cache[cacheKey] = bucket
 			}
-
-			actualDigest := reflistDigest(bucket)
-			if !bytes.Equal(actualDigest, digest) {
-				delete(collection.cache, cacheKey)
-				return fmt.Errorf("corrupt reflist bucket %d: expected digest %s, got %s",
-					idx,
-					bucketDigestEncoding.EncodeToString(digest),
-					bucketDigestEncoding.EncodeToString(actualDigest))
-			}
 		}
 
 		sl.bucketRefs[idx] = bucket
```

PR #1624, opened 2026-08-26
api/publish: allow setting custom Codename on the Release file

chipflake:
## Description of the Change

Currently, when publishing via the API, the Codename is always set to the same string as the Distribution.  This does not match the `aptly` CLI behaviour, where it is possible to set a `-codename` flag.

The CLI does not support updating the Codename after the published repo has been created (on Update), so I did not add `Codename` to the Update params here either. I think that it is ok not to support Codename updates (the Distribution cannot be changed either), but let me know if you think otherwise.

We are already using this patch internally, it would be nice to have it merged so that we can use upstream builds.

## Checklist

- [ ] documentation updated

### DIFF
--- a/api/publish.go
+++ b/api/publish.go
@@ -160,6 +160,8 @@ type publishedRepoCreateParams struct {
 	Sources []sourceParams `binding:"required"    json:"Sources"`
 	// Distribution name, if missing Aptly would try to guess from sources
 	Distribution string `                         json:"Distribution"          example:"bookworm"`
+	// Value of Codename: field in published repository stanza, if missing equals Distribution
+	Codename string `                             json:"Codename"              example:"bookworm"`
 	// Value of Label: field in published repository stanza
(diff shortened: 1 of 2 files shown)


saj:
> We are already using this patch internally

I would like to provide additional context on why this patch is helpful.
(Beyond improving API parity with the CLI, as noted by chipflake, above.)

```
  -codename="": codename to publish (defaults to distribution)
```

Since #1269, aptly now allows users to publish to suites (distributions) whose literal values include slashes (`/`).  We use this feature to publish to ornate suite names that have meaning to our internal org.  While we are happy to see these strings in our suite names, we don't think it is proper for the same literals to appear in `Codename:`.

This patch allows us to emit `Codename: trixie` or `Codename: bookworm` -- decoupling from our ornate suite names -- which is more in-keeping with upstream custom.

PR #1637, opened 2026-09-10
api: add metrics-only listener

luanmdang:
## Motivation

Aptly currently exposes Prometheus metrics through the REST API listener. Operators who need a loopback or monitoring-only scrape address must expose the API on that address too.

## Changes

- Add an optional CLI-only `-metrics-listen=host:port` flag to `aptly api serve`.
- Serve only exact `GET` and `HEAD` requests to `/metrics` on the dedicated plain-HTTP TCP listener.
- Require `enableMetricsEndpoint` and preserve the existing `/api/metrics` endpoint, including `HEAD` support.
- Pre-bind both listeners, stop the sibling if either server fails, and drain both concurrently with a bounded graceful shutdown before waiting for background tasks.
- Keep systemd socket activation scoped to the API listener while binding the metrics listener separately.
- Update command help, shell completions, the man page, and contributor metadata.
- Add focused routing, startup, failure, shutdown, race, and end-to-end system coverage.

The listener intentionally provides no TLS or authentication; operators can apply those controls through network policy or a proxy.

## Testing

- `go test ./cmd -count=10`
- `go test -race ./cmd`
- `go test -race ./api -check.f APISuite.TestHeadMetrics`
- `go vet ./cmd ./api`
- `go mod verify`
- pinned `golangci-lint` and `flake8` checks
- full Docker unit suite
- full Docker system suite, split into shards

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -84,3 +84,4 @@ List of contributors, in chronological order:
 * Zhang Xiao (https://github.com/xzhang1)
 * Tom Nguyen (https://github.com/lecafard)
 * Philip Cramer (https://github.com/PhilipCramer)
+* Luan Dang (https://github.com/luanmdang)
--- a/api/api_test.go
+++ b/api/api_test.go
@@ -158,6 +158,13 @@ func (s *APISuite) TestGetMetrics(c *C) {
 	c.Check(b, Matches, ".*aptly_build_info.*version=\"testVersion\".*")
 }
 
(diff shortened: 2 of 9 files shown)


luanmdang:
The remaining failures are all Debian 11/Bullseye jobs. Bullseye LTS ended 8-31, and its security repo metadata expired 9-7, so apt-get update fails before the project builds. Looks like all other lint and coverage build passes!

